PR #1, opened 2026-08-22
feat: add native Linux support, window controls, and UI polish

sahilcodexx:
## What changed

Adds clean, scoped Linux support and UI polish while preserving macOS chrome, vibrancy, and PATH resolution with 100% fidelity:

- **Clean Scoping & Merged Config**: Moved all Linux-specific window and bundle overrides into `src-tauri/tauri.linux.conf.json` (merged automatically by Tauri on Linux), leaving `src-tauri/tauri.conf.json` 100% untouched for macOS.
- **Window Controls & ACL Permissions**: Added explicit window ACL permissions (`allow-minimize`, `allow-toggle-maximize`, `allow-set-title`) in `src-tauri/capabilities/default.json` so native Minimize, Maximize/Restore, and `setTitle()` function properly at runtime.
- **Centralized Platform Utilities (`src/lib/platform.ts`)**: Deduplicated platform checks and shortcut symbols (`MOD`, `ALT`, `SHIFT`, `IS_MAC`) across all frontend surfaces into a single shared module.
- **Chrome & Tab Layout Preservation**: 
  - Restored tab overflow `useLayoutEffect` on tab list modifications in `TitleBar.tsx`.
  - Reverted sidebar header height back to `h-9.75` for exact 40px alignment with `TAB_BAR_HEIGHT` and the titlebar.
  - Centered system title in the title bar is gated to non-macOS (`!IS_MAC`), leaving the macOS titlebar visually identical to `main`.
- **PATH Precedence Preserved**: Kept `/opt/homebrew/bin` and `/usr/local/bin` ahead of system paths in `harness.rs` and `pty.rs` to maintain macOS Homebrew resolution, while appending Linux package directories (`/usr/bin`, `/snap/bin`, `~/.cargo/bin`, `~/.npm-global/bin`).
- **Alt Menu Bar Polish**: Removed inert `Edit` menu items and added a window `blur` listener to prevent Alt-tap arming from persisting across Alt+Tab window switches.
- **CI & Build Pipeline**: Added `ubuntu-latest` workflow matrix with WebKitGTK prerequisites and preserved `createUpdaterArtifacts: true`.

## Checklist

- [x] I ran `npm run check` (Vitest 35/35 test files, 330 unit tests passed, `tsc --noEmit` 0 errors, `cargo fmt`, `cargo clippy`, `cargo test` 93/93 tests passed)
- [x] macOS chrome, traffic lights, and window vibrancy verified untouched


hardbeat920:
Thanks for this! it's a lot of work, and the Linux-specific parts are mostly right. 
I ran it on macOS to check the cross-platform claims, and the core issue is scoping: several changes that need to be Linux-only are applied globally, and they regress macOS. The fixes are small.

## Blockers

**1. `decorations: false` is global and breaks macOS chrome** - `src-tauri/tauri.conf.json:22`

I verified this locally. On `main`, the traffic lights sit inline in the dark 40px tab bar as one unified header. With this change, macOS draws a solid light-gray native titlebar strip *above* the app's tab bar and pushes the whole UI down - very close to the white-headerbar artifact this PR fixes on Linux, just on the other platform. I isolated it: reverting `backgroundColor` while leaving `decorations: false` still reproduces it, so this line is the sole cause.

It's also redundant - `src-tauri/src/lib.rs:93` already calls `window.set_decorations(false)` under `#[cfg(not(target_os = "macos"))]`. Deleting the config key should be enough, or move it to a `tauri.linux.conf.json` (Tauri merges that automatically).
1-baseline-main

2-pr-both-lines

**2. `backgroundColor: "#17171a"` is global** - `tauri.conf.json:26`

The `#00000003` it replaces is deliberate; the header comment in `src-tauri/src/macos.rs` explains that a fully clear window plus a native shadow makes macOS draw a chamfered gap at the corners, and the tiny alpha avoids that. Not the cause of #1, but it should still be Linux-scoped. If you go the `tauri.linux.conf.json` route, `transparent` probably wants to be `false` there too, since that's the underlying WebKitGTK problem.

**3. The new window controls are likely non-functional except Close** - `src-tauri/capabilities/default.json:10`

`core:window:default` is already included by `core:default` on the line above, so adding it changes nothing. And it only grants getters plus `allow-internal-toggle-maximize`. The three new calls map to `plugin:window|minimize`, `plugin:window|toggle_maximize`, and `plugin:window|set_title`, none of which are granted - so Minimize, Maximize/Restore, the title-bar double-click handler, and the `setTitle()` sync should all reject at runtime. Close works because `allow-close` is listed explicitly.

Because `WindowControls.tsx` wraps `void getCurrentWindow().minimize()` in `try {}`, the rejection is asynchronous and never caught - it goes to the webview console and the button just does nothing, which would be easy to miss in testing. `document.title` still updates, so the in-app centered title looks correct even though the OS-level window title isn't being set.

Replace the added line with:

```json
"core:window:allow-minimize",
"core:window:allow-toggle-maximize",
"core:window:allow-set-title",
```

Worth confirming on your machine - if the buttons *do* work for you, I'd like to know what I'm missing in the ACL.

**4. `createUpdaterArtifacts: false`** - `tauri.conf.json:44`

This breaks the release/updater pipeline. For unsigned local bundles, override it at build time instead: `npm run tauri build -- --bundles deb --config '{"bundle":{"createUpdaterArtifacts":false}}'`.

**5. A tab-overflow effect was dropped** - `src/chrome/TitleBar.tsx`

`useLayoutEffect(() => { syncTabOverflow(); }, [activeId, collapsedGroups, syncTabOverflow, tabs])` is gone; only the ResizeObserver one survived. Tab overflow state goes stale on tab add/close/group-collapse, on both platforms. Looks like a rebase casualty rather than intentional.

**6. PATH reordering changes macOS resolution** - `src-tauri/src/harness.rs:927` and the same block in `pty.rs`

`/opt/homebrew/bin` moved from first to last, behind `/usr/local/bin`, `/usr/bin`, and `/bin`. On Apple Silicon that flips which `node`/`claude` wins for anyone with both a Homebrew and a system install. Please append the Linux directories rather than reordering the existing ones.

## Should fix

**7. Sidebar header is now 1px taller** - `src/chrome/Sidebar.tsx:279`

`h-9.75` + the tablist's `border-y` came to exactly 40px, matching `TAB_BAR_HEIGHT` in `macos.rs` and the title bar's `h-10`. `h-10` + `border-b` makes it 41px, so the sidebar chrome no longer lines up with the title bar or the pinned traffic lights on macOS.

**8. Platform detection is duplicated nine times.** Eight modules each define their own `IS_MAC`, plus `index.html` and `appearance.ts`. All use the deprecated `navigator.platform`. One `src/lib/platform.ts` export would do it - and since `initAppearance()` already sets the class, components could read that instead. The `is-non-mac` class is set in two places and never used; the CSS uses `html:not(.is-mac)`.

**9. The Edit menu is mostly inert** - `src/chrome/MenuBar.tsx`. `document.execCommand("paste")` is a no-op in WebKitGTK, and undo/redo/cut/copy won't reach xterm or CodeMirror, which manage their own state. I'd either wire these to the real handlers or drop the Edit submenu rather than ship menu items that silently do nothing.

**10. `html:not(.is-mac) [class*="backdrop-blur-"]` with `!important`** - `src/index.css:1223`. A substring match on Tailwind class names will capture every blur surface added from here on, intended or not.

**11. Unrelated changes.** The seven new direct deps, `unified`, and the vitest `dedupe`/`inline` config aren't part of Linux support. Promoting transitive deps to direct is reasonable, but it belongs in its own PR.

**12. The centered window title renders on macOS too.** That's a visual change to mac beyond the scope of Linux support - worth gating on `!IS_MAC` or raising separately.

**13. Alt-tap arming survives focus loss** - `MenuBar.tsx`. Alt+Tab away and back and `altPressedAlone` is still `true`, so the next bare Alt keyup opens the menu unexpectedly. A `blur` listener that resets the flag fixes it.

## What's good

CLI discovery paths, the `$SHELL` fallback with the new `is_file()` check, the `xdg-open` branch in `reveal_path`, the `app.set_menu` scoping, the `WindowControls` markup itself, the CI matrix, and the docs updates all look right to me.

I also checked the shortcut relabeling specifically, since `Ctrl+` labels would be misleading if the handlers only accepted Cmd - they don't. Everything goes through `mod = metaKey || ctrlKey`, so the new labels are accurate on Linux. Nice catch on `apply_gui_path` picking up `~/.cargo/bin` too.

Most of the blockers collapse into one change: move the window config into `tauri.linux.conf.json` and leave `tauri.conf.json` alone.

Happy to look again once you've pushed.

sahilcodexx:
Thanks for the thorough review and screenshots! All 13 points have been addressed in the updated branch:

### Blockers Addressed
1 & 2. **Config Scoping**: Reverted `src-tauri/tauri.conf.json` to match `main` completely. Moved Linux-specific window properties (`decorations: false`, `transparent: false`, `backgroundColor: "#17171a"`, `bundleMediaFramework: false`) into `src-tauri/tauri.linux.conf.json` (auto-merged by Tauri on Linux).
3. **Window ACLs**: Added `core:window:allow-minimize`, `core:window:allow-toggle-maximize`, and `core:window:allow-set-title` to `src-tauri/capabilities/default.json`.
4. **Release Pipeline**: Restored `createUpdaterArtifacts: true` in `tauri.conf.json`.
5. **Tab Overflow**: Restored `useLayoutEffect` for `syncTabOverflow()` in `src/chrome/TitleBar.tsx`.
6. **PATH Precedence**: Restored `/opt/homebrew/bin` and `/usr/local/bin` at the front of `PATH` in both `harness.rs` and `pty.rs` before appending Linux paths.

### Polish & Cleanups Addressed
7. **Sidebar Height**: Reverted header back to `h-9.75` (exact 40px match with `TAB_BAR_HEIGHT`).
8. **Platform Utilities**: Created `src/lib/platform.ts` with `IS_MAC`, `MOD`, `ALT`, `SHIFT` and removed duplicate declarations across 9 files.
9. **Menu Bar Edit Items**: Dropped the inert `Edit` menu from `MenuBar.tsx` (retained working File, View, Terminal actions).
10. **CSS Polish**: Removed the broad `[class*="backdrop-blur-"]` wildcard selector from `index.css`.
11. **Scoped Diff**: Reverted unrelated package promotions in `package.json` and vitest config.
12. **macOS Drag Region**: Gated the centered title bar text to `!IS_MAC` so macOS titlebar visual layout is identical to `main`.
13. **Alt Key Focus**: Added a `window` blur listener in `MenuBar.tsx` to disarm `altPressedAlone` on focus loss / Alt+Tab.

All check suites (`npm run check`) pass cleanly with 0 errors across Vitest (330 tests), TypeScript (`tsc --noEmit`), `cargo fmt`, `cargo clippy`, and `cargo test` (93 tests).

hardbeat920:
This is a big improvement! 

I checked every item from the last round and they're all addressed, and folding the pre-existing duplicate platform checks in `FileTree`/`DiffPane`/`TerminalView` into `platform.ts` was more than I asked for.

I ran the branch on macOS - traffic lights are back inline in the tab bar, no gray strip above the chrome, no centered title:

3-macos-chrome-verified

The title isn't visible there because macOS uses `hiddenTitle`, but querying the window directly is getting through and the capability fix works. That blocker is closed.

Two things left.

**`tauri.linux.conf.json` drops the rest of the window config.** Tauri merges platform configs with JSON Merge Patch, which replaces arrays instead of merging them element-wise - so `app.windows` on Linux becomes only what's restated in that file. I confirmed it by dropping an equivalent `tauri.macos.conf.json` into my tree: the window title came up as `Tauri App` instead of `MonoCode`, and `MonoCode` again with the file removed.

So Linux is currently losing `title`, `width`, `height`, `minWidth`, `minHeight`, `backgroundThrottling`, and `shadow`. A fresh install opens at Tauri's default 800×600, and the window can be resized below the layout's minimum. It's easy to miss because the window-state plugin restores geometry for anyone who's already run it, and the new `setTitle()` overwrites the title once the frontend mounts. Fix is just to restate the full window object in the Linux file.

**The sidebar divider went missing.** The header is back to `h-9.75`, but the tablist is still `border-b` from the last round - that top border was the line between the header and the Sessions/Files row. Reverting it to `border-y` should do it. `main` vs branch:

1-sidebar-divider-main

2-sidebar-divider-branch

Non-blocking, take or leave: `editorSearch.ts` renders `Alt+Ctrl+F` and `Shift+Ctrl+G` while `MenuBar` renders `Ctrl+Shift+N` - worth picking one order. And in `WindowControls`, if the component unmounts before `onResized` resolves, `unlisten` never gets assigned and the listener leaks.

If you can fix those two I'm happy to merge. I'll re-run the macOS check on the next push.
Thank you

sahilcodexx:
Thanks again! I've addressed those final items in the latest commit:

1. **Complete Linux Window Config**: Restated the full window object in `src-tauri/tauri.linux.conf.json` (`title`, `width`, `height`, `minWidth`, `minHeight`, `backgroundThrottling`, `shadow`, alongside Linux `decorations: false`, `transparent: false`, `backgroundColor: "#17171a"`) to ensure JSON Merge Patch doesn't drop default window dimensions and geometry limits.
2. **Sidebar Divider Restored**: Changed the tablist border back to `border-y border-content/10` in `src/chrome/Sidebar.tsx` to restore the divider line between the header and the Sessions/Files row.
3. **Consistent Shortcut Modifiers**: Standardized modifier order in `src/surfaces/editorSearch.ts` with `MOD` first (`${MOD}${ALT}F` and `${MOD}${SHIFT}G`) to match `MenuBar`.
4. **WindowControls Listener Cleanup**: Guarded the `onResized` subscription with a mounted flag in `src/chrome/WindowControls.tsx` to immediately unlisten if the component unmounts before the promise resolves.

All checks pass cleanly (`npm run check` - 330 Vitest tests, `tsc --noEmit` 0 errors, `cargo clippy` 0 warnings, 93 Rust tests).

hardbeat920:
@sahilcodexx Thank you. This looks great!

PR #14, opened 2026-08-26
Allow quoting of selected text from agent’s response

kinsomicrote:
## What changed

Add an option to quote selected text from agent's response into message

## Why

I recently started try monocode after using bb, I liked this feature when using bb and thought it will be a good addition

## UI

CleanShot 2026-08-26 at 16 59 58

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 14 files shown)


kinsomicrote:
Hello @hardbeat920, thanks for building Monocode. I'm not sure if this is in line with the direction you have for the app. Let me know if you have any thoughts. Thanks.

hardbeat920:
I love it. Will review tomorrow but from the screen recording looks great

hardbeat920:
Thanks for the great work @kinsomicrote 
This landed in 0.1.11

PR #15, opened 2026-08-27
Preserve streamed text across provider harnesses

kinsomicrote:
## What changed

- Appends each text delta exactly once.
- Keeps spaces, blank lines, and repeated characters.
- Preserves the order of text and tool calls.

## Why

The PR resolves the text streaming issue as shown in the images below

CleanShot 2026-08-27 at 07 18 02@2x

## UI

CleanShot 2026-08-27 at 13 34 42@2x

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Streamed responses preserve repeated letters, punctuation, spaces, blank lines, and Markdown boundaries across supported providers.
+
 ## [0.1.13] - 2026-08-27
 
 ### Added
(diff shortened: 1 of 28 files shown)


kinsomicrote:
Hello @hardbeat920 

This resolve the formatting issue I noticed. I could only test using Pi and Codex, I also did not scope it down to just Pi (which is the coding agent I use) as it would meant introducing a new event type to `HarnessEvent`. Let me know if there are things you'll need me to change.

hardbeat920:
@kinsomicrote yeah this is a real bug. I will review and let you know 🙏

hardbeat920:
mergeStream overlap-matching plus dropping whitespace deltas is what smashed headings/tables. Thank you for catching that and for the tests around exact text.

I want that fix, but I’d rather keep one join apply.ts so every harness provider stays identical. I landed a smaller cut of the same diagnosis here f61fc2f

I’m going to close this PR in favor of that smaller change - not because the bug wasn’t real.

Happy to review a follow-up.

PR #22, opened 2026-08-28
Make the tab-closing project aware

kinsomicrote:
## What changed

When closing tabs now, the next tab is scoped to the tabs in that project rather than to global tabs.

## Why

Say you have two projects, monocode and ruler. Then you create a session: monocode, ruler, then monocode. Closing the tab for the second session in monocode moves the user to the tab in the ruler project. 

## UI

CleanShot 2026-08-28 at 18 40 41

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Closing a tab in Deck keeps the active tab in the current project when another tab from that project is open.
 - Streamed Markdown keeps headings, blank lines, tables, and repeated characters. Completed Claude/Codex snapshots no longer paste the same reply twice. Diagnosed in #15 by @kinsomicrote.
 
 ## [0.1.13] - 2026-08-27
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -177,7 +177,12 @@ import {
   rememberProject,
   sameProjectPath,
 } from "./lib/recents";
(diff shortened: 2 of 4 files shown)


hardbeat920:
@kinsomicrote yeah this is a real bug. thank you for catching it.

in deck, everything is already scoped to the current project — the tab strip, next/prev, jumping by number. close still uses the global tab list, so you can be in monocode, close a monocode tab, and land in another project. that's the issue.

classic is the other way around: all tabs from every project sit in one strip, so close should keep picking a neighbor from that full list.

could you keep the project-aware next-tab for deck only, and leave classic as it is? the planner idea is good. we don't need to rewrite the rest of close / last-tab reset / `onClearTabSession` for this — just which tab gets activated after a close when deck is on.

i'll wait on you to update so you get the credit, rather than landing a smaller cut myself.

hardbeat920:
@kinsomicrote yeah this is a real bug. thank you for catching it.

in deck, everything is already scoped to the current project. classic is the other way around: all tabs from every project sit in one strip, so close should keep picking a neighbor from that full list.

Could you keep the project-aware next-tab for deck only, and leave classic as it is? 
the planner idea is good. we don't need to rewrite the rest of close / last-tab reset / onClearTabSession for this, just which tab gets activated after a close when deck is on.

i can wait for you or I can land a fix myself...?

hardbeat920:
@kinsomicrote this is the one. I tested in deck: two tabs in the same project, close one, stays in that project. classic still uses the full strip.

merging this now. thank you.

PR #23, opened 2026-08-28
Report failed Pi turns and meter their context

emircan-sahin:
## What changed

A Pi or omp turn that fails now reports the reason instead of ending silently, and the context meter follows the turn instead of sitting at zero until it is over.

## Why

Pi does not send an error frame when a turn fails. It ends the assistant message with `stopReason: "error"` and the reason in `errorMessage`, and the adapter read neither field, so `settleTurn` closed the turn with no text and no error. On a machine whose provider token had expired, every Pi turn came back empty and it looked like the agent was ignoring the prompt — the CLI had been saying `No API key for provider: openai-codex` the whole time.

The failure is held until `agent_end` reports the turn is not being retried, so an auto-retry does not surface the attempt it replaces.

`contextFromUsage` had the same shape problem in reverse: it only looked at `rec.usage`, the top of a frame, which Pi never populates. A finished message carries usage on `message`, a streaming one on the partial inside `assistantMessageEvent`. The meter therefore stayed at zero for the whole turn and only caught up from `get_session_stats` once it settled.

Codex already maps a failed turn to `session.error`; this brings Pi in line.

## UI

No layout change. A failed turn now renders the harness error the transcript already has a place for, where it previously rendered nothing.

## Notes

Frame shapes were captured from a live `pi --mode rpc` turn on `claude-sonnet-4-6` (pi 0.80.2) and the new tests assert against those exact shapes, including the `message_end` / `message_update` split and the user-message frame that must stay `null`.

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- A Pi or omp turn that fails now says why. Pi reports a failure inside the assistant message rather than as an error, so an expired provider token ended the turn with no reply and no explanation — it looked like the agent was ignoring you.
+- The context meter stayed at zero for the whole of a Pi or omp turn. Usage was only read from the top of a frame, where Pi never puts it; the meter now follows the turn instead of catching up once it ends.
+
 ## [0.1.16] - 2026-08-28
 
 ### Added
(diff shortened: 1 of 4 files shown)


hardbeat920:
Landed on main with credit @emircan-sahin! thanks for this 🙏 

Same diagnosis: Pi reports a failed turn on the assistant message (stopReason: "error" / errorMessage) rather than an error frame, so an expired token looked like the agent ignoring the prompt. 

I added a guard so tool-result usage cannot overwrite the context meter (nested LLM calls are not the window level).

Changelog credits you in #23.

PR #24, opened 2026-08-28
Read model catalogs on demand instead of at launch

### LINKED ISSUE #19: The Pi agent stays running in the background at all times
### MonoCode version

latest

### macOS version

26.4.1

### Provider CLI

Multiple / not sure

### What happened

I was only using Codex and Claude Code, but when I checked Activity Monitor, Pi agent was eating up 1GB of RAM even though I wasn't using it at all.

Unused agents shouldn't be running in the background at all, they should only start up when I actually want to use them.

### Steps to reproduce

Open the app and check Activity Monitor right away.

emircan-sahin:
## What changed

Model catalogs are read on demand instead of at launch, and the one-shot probe children now have an owner on the Rust side that outlives the webview.

Closes #19.

## Why

`refreshHarnessCatalogs()` ran on app mount and called every registered adapter, so each installed provider CLI was started at launch purely to list its models. Someone using only Claude and Codex still got a `pi` process in Activity Monitor, which is exactly what #19 reports.

Catalogs are now read for the providers that open sessions already use. The rest wait until the user asks: their tab in the model picker, or Settings → Providers, which is the one page that lists every provider's models and so is the right place to pay for the probes. This generalises the lazy refresh the codex tab already did for itself, and `refreshHarnessCatalog` keeps a per-run set so a probe does not repeat — while a failed probe is dropped from that set so a CLI that was missing or busy gets another chance.

The probes also had no owner. Cleanup was a `finally` in the webview, which never runs if the window reloads or closes mid-probe; the leaked `pi` probe in #19 had reached 1 GB and 14 minutes of CPU. One-shot children now pass a `ttl_ms` that `harness_spawn` enforces from a watchdog thread. It only fires while the registry entry is still the child it spawned, so a normal kill or a respawn wins the race. The wait thread matches on pid for the same reason: shared ids (`monocode-*-probe`, `monocode-*-text`) get respawned under one key, and a dying generation must not evict the child that replaced it — that entry is the only handle `kill_all` has left.

Listing Pi's models needs no extension host either. Loading one cost 2.9s and 357 MB against 0.8s and 190 MB, and it pulled in the child process that kept the leaked probe busy. The model list is identical either way.

## UI

No visible change. The model picker fills its tab on open the way the codex tab already did; the CLI availability probe still runs at startup, so the installed/not-installed hints are unaffected.

## Notes

`a_dead_generation_does_not_evict_its_replacement` covers the pid-matching rule with a real stdin handle on a dead process, so a fake pid can never reach `terminate`.

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Model catalogs are read on demand instead of at launch. MonoCode used to start every installed provider CLI on startup just to list its models, so an agent you never opened still showed up in Activity Monitor. Now only the providers your open sessions use are read at launch; the rest are read when you open their tab in the model picker, or when you open Settings → Providers. ([#19](https://github.com/hardbeat920/monocode/issues/19))
+
+### Fixed
+
+- A model-catalog or usage probe left behind by a window reload or close is now killed by the backend instead of living on. One stranded `pi` probe had reached 1 GB and 14 minutes of CPU. ([#19](https://github.com/hardbeat920/monocode/issues/19))
+- Listing Pi's models no longer loads your extensions: 0.8s and 190 MB instead of 2.9s and 357 MB, and one fewer child process.
+
 ## [0.1.16] - 2026-08-28
 
 ### Added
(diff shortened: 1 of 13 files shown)


hardbeat920:
Thanks @emircan-sahin same diagnosis as #19. The lazy catalogs / skip-Pi-extensions fix already landed on main in b26f12c, so I’m closing this as superseded. 

If you want to split out the Rust ttl_ms backstop for a probe that survives a webview reload, that part is still missing and I’d take a small PR for it.

PR #25, opened 2026-08-28
Run Claude Code hooks in interactive sessions

emircan-sahin:
## What changed

Claude Code hooks configured in the user's `settings.json` now run in interactive sessions, with a Settings → General toggle to turn them back off.

## Why

MonoCode launched the CLI with `disableAllHooks`, so every hook a user had configured — `PreToolUse` command rewrites, blocks, notifications — was silently skipped. Nothing surfaced that they had been dropped, so a hook that worked in the terminal simply did nothing in MonoCode.

Hooks are now disabled only for isolated spawns, which are MonoCode's own helper calls such as title and summary generation. Those should not fire a user's hooks, and they still do not.

The toggle (on by default) is the escape hatch for a hook that misbehaves badly enough to need the session back without editing `settings.json`.

Turning hooks on exposed a second bug: a `PermissionRequest` hook can resolve a prompt before the user does, and the transcript labelled that "Rejected". The approval decision gains a `cancelled` state so the transcript stops blaming the user for a decision a hook made.

## UI

Settings → General gains a "Claude Code hooks" row. A permission prompt resolved by a hook no longer shows "Rejected".

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Claude Code hooks now run. MonoCode used to launch the CLI with `disableAllHooks`, so every hook in your `settings.json` — command rewrites, blocks, notifications — was silently skipped. Settings → General has a "Claude Code hooks" toggle (on by default) to turn them back off if one misbehaves. MonoCode's own helper spawns, like title generation, stay hook-free.
+
+### Changed
+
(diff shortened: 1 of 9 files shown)


emircan-sahin:
I think this PR is urgent. Without hooks, the models burn through way too many tokens and get things done way slower. I've also got some skills that improve code quality, but since hooks are disabled, nobody can actually use them.

@hardbeat920

hardbeat920:
Thanks @emircan-sahin!

`disableAllHooks` was intentional. `PermissionRequest` hooks run in parallel with the stdio prompt and dismiss it when they allow/deny, and MonoCode is the approval UI, so we turned every hook off rather than let them race the toast. 

That was broader than it needed to be.

Issue #20, opened 2026-08-28
Usage limit section refresh frequency and bugs in the UI


emircan-sahin:
### MonoCode version

latest

### macOS version

26.4.1

### Provider CLI

Claude Code

### What happened

When you're using Claude Code, the footer shows a detailed percentage of how much you have left. To get this, the app pulls data from Claude Code every 5 minutes, but the UI itself only refreshes every 15 minutes.

Users might want to see this updated faster, after all, you already have the data on hand. You could also pull this data every time a session finishes, meaning whenever a message comes back from the AI, and refresh the UI right then too.

Also, if you open the app while offline and your internet connection comes back afterward, the usage limit should refresh. (Maybe also when a 401 error is received.)

### Steps to reproduce

Open Claude Code, give it a few tasks, and wait a bit.

hardbeat920:
@emircan-sahin thanks for this!

The usage chips don’t fetch every 5 minutes and then wait 15 minutes to update. There’s one fetch path, and the footer paints as soon as that fetch returns. The two timers do different jobs:

- 15 minutes is the background poll while the window is visible
- 5 minutes is only a freshness check: if you refocus the app, we skip another request when the last snapshot is still newer than that
- The countdown ticks on its own, the percent used only changes after a new fetch. 
 
There’s also a refresh button on the footer if you want the latest numbers right away.

So the “we already have the data, the UI is just stale” part doesn’t look like a bug. 
A 401 on the usage request is already retried after an OAuth refresh, if that still fails, the chip shows expired.

I am open to discuss as i might be wrong in my evaluation

emircan-sahin:
Hi, you might be right about this. Also, what if we update the remaining usage limit shown in the UI whenever a session gets a response from the AI, meaning whenever the AI sends a message?

I opened a PR for this, but it needs review.

PR #26, opened 2026-08-28
Refresh usage footer on turn end and stop rotating Claude's OAuth token

emircan-sahin:
## What changed

The usage footer refreshes when a turn ends and polls on a cadence that can actually deliver a fresh snapshot, plus four related fixes.

Addresses #20.

## Why

#20 is right that the data was already on hand: the fetch floor was five minutes but the UI timer ticked every fifteen, so a snapshot could sit unused for ten minutes. The timer now ticks every minute and a snapshot lands as soon as it is eligible, and a finished Claude or Codex turn marks the provider stale so the number moves when the user has just spent some. Both providers keep the five-minute floor between reads.

Also in #20's list, a provider that was not signed in at launch stayed "not connected" for the life of the process; it is now retried every 15 minutes so signing in later recovers on its own. Clicking refresh during a background poll did nothing at all — the click is now queued and runs right after. A 429 backs off for 30 minutes rather than retrying on the normal cadence, with the reason in the chip's tooltip.

The footer no longer refreshes Claude Code's OAuth token. Those credentials belong to the `claude` CLI, which MonoCode already spawns for every turn and which rotates them itself. Writing to its keychain entry meant a rotation MonoCode could not save left the CLI holding a dead refresh token. The footer is now a read-only view: an expired token shows `expired` until the next Claude turn renews it.

One unrelated-looking line rides along because it is the same polling story: an idle terminal forked a `ps` every second per pane to name its foreground process, and now bails before the fork when the foreground pid is the shell itself.

## UI

The footer's number updates sooner. A throttled provider shows the reason in the chip tooltip; an expired token shows `expired`.

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- The usage footer refreshes as soon as a Claude or Codex turn ends, and the poll timer now ticks every minute so a snapshot lands as soon as it is eligible instead of up to fifteen minutes late. Both providers keep a five-minute floor between reads.
+
+### Fixed
+
+- A provider that was not signed in when the app launched no longer stays "not connected" forever; it is retried every 15 minutes, so signing in later recovers on its own.
+- Clicking refresh while a background poll was running did nothing. The click is now queued and runs right after.
+- An idle terminal no longer forks a `ps` every second to name its foreground process.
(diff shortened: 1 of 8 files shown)


nulljosh (commented):
Solid change and good rationale — dropping the token-refresh logic here makes sense (MonoCode writing a rotated token to the CLI's keychain entry while the CLI itself also refreshes independently is a real race that can dead-end the refresh token), and read-only is the safer contract. The pty.rs foreground_label short-circuit and the 429 backoff/message are also correct, minimal fixes.

This PR currently has a merge conflict against main and can't be merged as-is — needs a rebase from the author.

PR #27, opened 2026-08-28
Scope session changes to the session's own baseline

emircan-sahin:
## What changed

The Changes panel gains a Project/Session toggle, and the session view now measures a file against the snapshot taken when the session started rather than against git's HEAD.

## Why

The session file list took its line counts from git's HEAD numbers. A file that was already dirty when the session started therefore billed someone else's edits to that session, and opening it diffed against the index instead of the snapshot. The number and the diff both described the working tree, not the session.

`session_change()` now decides in one place whether a path changed and which baseline applies, and carries the contents it read, so a status poll reads each file once instead of twice. A file with a snapshot gets its counts from a bounded line diff against that snapshot; a file that was clean at session start keeps git's numbers, since HEAD is its baseline.

`session_checkpoint_file_diff` is the command behind the editor diffing a file against the session snapshot. Session review tabs are keyed separately from project review tabs so the two views of one file are two tabs rather than one that flips meaning.

## UI

The Changes panel gains a Project/Session toggle and a session-scoped file list with per-file Keep and Undo.

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- The Changes panel has a Project/Session toggle. The session view lists only the files that session touched, each with its own Keep and Undo, and opening one diffs it against the snapshot taken when the session started instead of against the index. A session review tab and a project review tab of the same file are now separate tabs.
+
+### Fixed
+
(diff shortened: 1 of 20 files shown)


emircan-sahin:
Closing this: the session-scoped baseline turned out to be buggy in practice — a session with 8 changed files on disk was reported as only 1 changed file in the panel. Reverted from the local branch as well; reopening once the baseline tracking is fixed.

PR #28, opened 2026-08-28
Keep package-lock.json in sync on a version bump

emircan-sahin:
## What changed

`npm run set-version` now updates `package-lock.json`, and the lockfile is brought back in sync.

## Why

The bump script updates `package.json`, `Cargo.toml`, `tauri.conf.json` and `Cargo.lock`, but never `package-lock.json` — so the lockfile has called itself 0.1.0 since the first release, sixteen versions ago, and its root dependency list had drifted from `package.json`.

The lockfile names its version twice, at the top and on the root package entry. The top-level object repeats the same `name`/`version` pair as the root package, so the second pattern is anchored on `packages` to avoid matching the first one again. Verified by bumping to a throwaway version and checking both fields moved, then restoring.

`npm ci` tolerated the drift, so this is housekeeping rather than a broken build.

## UI

None.

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- `npm run set-version` left `package-lock.json` behind, so the lockfile still called itself 0.1.0 sixteen releases on. The script now updates both of the version fields it carries, and the lockfile is back in sync.
+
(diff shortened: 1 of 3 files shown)


hardbeat920:
Thanks @emircan-sahin  this is the right fix. 
set-version should have been updating both lockfile version fields all along!

I merged main into the branch so the changelog wouldn’t conflict. 

I’ll squash-merge once CI is green.

PR #31, opened 2026-08-28
Portal the model picker menu out of the composer

### LINKED ISSUE #29: Model can't be selected in sessions opened side-by-side or stacked
### MonoCode version

latest

### macOS version

-

### Provider CLI

Multiple / not sure

### What happened

Model can't be selected in sessions opened side-by-side or stacked. I can't select a model because the dropdown values are getting crushed behind the background due to z-index. The z-index needs to be increased to fix this.

Image

### Steps to reproduce

1. Open the app and create a session
2. Create another session next to it, then click on the model selector

emircan-sahin:
## What changed

The model picker's dropdown is now rendered through a `createPortal` onto `document.body` at the same `zIndex: 80` tier the other popups in the app use, instead of inline inside the composer.

## Why

Fixes #29.

`data-composer-box` in `Composer.tsx` carries `relative z-10`, which makes it a **stacking context**. The picker's menu is `position: fixed`, but it was rendered as a DOM descendant of that box, so its `zIndex: 50` resolved *inside* the box — against everything outside the composer the menu painted as if it were `z-10`.

Anything layered above that covered it. The case in the issue: an empty session in a split pane docks its composer outside `EmptySession`, while the arcade backdrop goes to `absolute inset-0 z-20` with an opaque `bg-background-base` (`TerminalGridBackground.tsx:350-356`). z-20 beats the menu's effective z-10, so the list was swallowed and no model could be selected.

Raising the inline z-index would not have helped — the menu has to leave the stacking context. The portal follows the pattern already used by `ExplorerMenu`, `TabGroupMenu`, `SessionFiltersMenu`, `SecondOpinionButton` and `TranscriptSelectionMenu`.

One adjustment came with it: the outside-click handler compared against the trigger's `root` ref, which no longer contains the portalled menu, so clicking inside the menu would have dismissed it. It now treats the menu node as inside too.

Positioning is unchanged — `menuStyle` is viewport-relative `position: fixed`, and nothing on `html`/`body`/`#root` creates a containing block for fixed elements.

## UI

Before: https://github.com/user-attachments/assets/9ba54eb8-48cf-4bfd-b1f2-702997d0804e

## Checklist

- [x] I ran `npm run check` (`check:web` — 694 tests, `tsc --noEmit` clean; the change is frontend-only and does not touch Rust)

## Note on scope

`ModelSettings`, `AccessPicker`, `BranchPicker` and `CwdPicker` sit in the same composer box with the same inline `position: fixed` + `zIndex: 50` and are hidden the same way. I kept this PR to the picker in the issue — happy to send the rest as a follow-up, or drop `z-10` from `data-composer-box` instead if you'd rather fix all five at the source.

### DIFF
--- a/src/chrome/ModelPicker.tsx
+++ b/src/chrome/ModelPicker.tsx
@@ -10,6 +10,7 @@ import {
   type KeyboardEvent as ReactKeyboardEvent,
   type ReactNode,
 } from "react";
+import { createPortal } from "react-dom";
 import {
   coerceModelPickerTab,
   findModel,
@@ -69,7 +70,7 @@ function menuStyle(anchor: DOMRect): CSSProperties {
     bottom: window.innerHeight - anchor.top + 6,
     width,
     height: Math.max(180, Math.min(340, anchor.top - 12)),
-    zIndex: 50,
+    zIndex: 80,
   };
 }
 
@@ -102,6 +103,7 @@ export function ModelPicker({
   const [favorites, setFavorites] = useState(loadFavoriteModels);
   const [menu, setMenu] = useState<CSSProperties>();
   const root = useRef<HTMLDivElement>(null);
+  const menuRef = useRef<HTMLDivElement>(null);
   const search = useRef<HTMLInputElement>(null);
   const onCloseRef = useRef(onClose);
   onCloseRef.current = onClose;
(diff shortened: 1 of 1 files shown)


hardbeat920:
Thanks @emircan-sahin  this is the right fix. Merging

PR #33, opened 2026-08-28
Carry the composer's model pick to new sessions

### LINKED ISSUE #30: Default model selection isn't working right!
### MonoCode version

latest

### macOS version

-

### Provider CLI

Claude Code

### What happened

Default model selection isn't working right!

I usually use Opus 5, but whenever I create a new session, it starts with Composer or Pi instead. If I've already updated the model or effort level in any chat, new sessions should inherit that same setting.

This would make the app feel faster and much more comfortable to use.

### Steps to reproduce

1. Create a session and check which model is selected.

emircan-sahin:
## What changed

`onModelChange` now calls `saveLastModelChoice(harness, resolved.id)`, so switching model from the composer's picker also updates the default new sessions start with.

## Why

Fixes #30.

`monocode.lastModel` was written only by Settings → Providers (`SettingsView.tsx`). The composer picker updated just the session it was opened from, so `defaultSessionChoice()` kept reading a stale (or absent) choice and fell back to the hardcoded `"cursor"` provider — every new session started on Composer regardless of what the user had actually been working in.

The effort/thinking chips sitting next to the picker already do the analogous thing (`onModelSettingsChange` → `saveLastModelSettings`), which is why effort was sticky and model was not. This just closes that gap. `saveLastModelChoice` also writes through to `defaultModels[harness]`, so the per-provider default in Settings tracks the picker too.

## UI

No visual change.

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- The model you pick in the composer now carries to new sessions. Previously only the Providers settings page wrote the default, so a session started from the built-in default no matter what you had switched to while working. Reported in #30.
 - Unused provider CLIs no longer start at launch. Catalog probes run only for harnesses in the restored workspace, or when you open that provider in the model picker. Pi/omp probes skip extensions so a leftover `pi` process cannot sit at ~1GB while you work in Codex or Claude. Diagnosed in #19.
 
 ## [0.1.16] - 2026-08-28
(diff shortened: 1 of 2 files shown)


hardbeat920:
Thanks for the contribution @emircan-sahin ! I’d like new sessions to keep using the default explicitly chosen in Settings. With this change, selecting a model for an individual chat silently replaces that default, which could be surprising.
I’m closing this for now. Remembering the last selection could still make sense as an optional behavior, but I’d keep it separate from the configured default.

PR #35, opened 2026-08-29
Allow monocode to fetch skills directly from pi

kinsomicrote:
## What changed

This allows monocode to load skills directly from pi

## Why

CleanShot 2026-08-29 at 14 32 46@2x

## UI

CleanShot 2026-08-29 at 14 33 40@2x

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -229,10 +229,10 @@ import {
   tabVisitForward,
   type TabVisitHistory,
 } from "./lib/tabVisitHistory";
-import { applySkillsToTurn } from "./lib/skills";
-import { applyFileMentionsToTurn } from "./lib/fileMentions";
+import { preparePrompt } from "./lib/promptPreparation";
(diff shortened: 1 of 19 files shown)


hardbeat920:
Thanks for this @kinsomicrote, the Pi catalog probe, caching, and process cleanup look solid.

One thing to fix before merge: 
the picker shows _/skill:architect_, but ranking still matches on the short name and the slash tokenizer rejects :. 
Typing skill empties the list. Search against invocation as well, and allow : in the slash token.

Let me know if this makes sense. Thanks again.

kinsomicrote:
CleanShot 2026-08-30 at 11 03 52@2x

CleanShot 2026-08-30 at 11 03 39@2x

@hardbeat920 does this match what you have in mind? I've pushed an update for it

hardbeat920:
This looks solid. Thanks @kinsomicrote

PR #36, opened 2026-08-30
Add a System theme option that follows the OS appearance

emircan-sahin:
The theme picker only offered Dark and Light, so the app never followed the OS appearance. This adds a third choice, System, which resolves through `prefers-color-scheme` and keeps tracking it while the app is open.

- `ColorScheme` stays the resolved value (`dark` | `light`); a new `ThemePreference` adds `system` and is what gets stored and shown in Settings.
- `applyThemePreference()` resolves the preference before toggling `theme-light`, so every existing consumer of `SCHEME_CHANGE_EVENT` (transcript, terminal, editor, mermaid) keeps receiving a concrete scheme and needs no change.
- `watchSystemColorScheme()` re-applies on OS flips, but only while the preference is `system`.
- The pre-paint script in `index.html` resolves `system` the same way, so the boot splash does not flash the wrong scheme before the bundle loads.
- The stored key is unchanged, so existing `dark`/`light` values keep working; unknown values fall back to the default. Default stays Dark, so nobody's current appearance changes.

Tests cover persistence of all three values, unknown-value fallback, system resolution in both directions, and the no-`matchMedia` fallback.

`npm run check:web` passes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -23,7 +23,13 @@
             var o = Math.max(0.15, Math.min(1, Number(opacity) || 0.85));
             root.style.setProperty("--sidebar-opacity", String(o));
           }
-          if (localStorage.getItem("monocode.colorScheme") === "light") {
+          var scheme = localStorage.getItem("monocode.colorScheme");
+          if (scheme === "system") {
+            scheme = window.matchMedia("(prefers-color-scheme: light)").matches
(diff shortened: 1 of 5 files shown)


hardbeat920:
Looks good, merging!

PR #37, opened 2026-08-30
Drag the window from any empty spot in the header

emircan-sahin:
Moving the window only worked from a few spots next to the window controls. With a project open the Workspace header — "Workspace" on the left, search and plus on the right, empty in between — could not be grabbed at all, and neither could the gaps in the tab strip.

The cause is the value of the attribute, not where it sits. Every header used the bare `data-tauri-drag-region`, and Tauri only starts a drag when the mousedown target *is* that element. The `Workspace` label is `flex-1`, so it covers the whole middle of that row: the area that looks empty is the label's own box, and it swallowed the gesture. The title bar's tab strip went further and set `data-tauri-drag-region="false"`, so its padding and the gaps between tabs never dragged either.

Tauri supports `"deep"`, which drags from anywhere in the subtree while still exempting buttons, links, inputs and anything marked `"false"`. That matches what a title bar should do, so every header row now uses it — the title bar, the sidebar, the project rail, and the inbox, notes, search and settings surfaces, which render their own title bar row and had the same dead spots on their icon and label.

- every gap, label and spacer in those rows drags the window
- the tab strip's gaps drag; tabs stay clickable and re-orderable, since Tauri exempts them as buttons
- the existing `data-tauri-drag-region="false"` opt-outs (pickers, window controls, drag handles, the search field) keep working untouched
- the per-child attributes on spacers and title spans are redundant under `"deep"` and were removed

The rows are also `select-none` now. Tauri's drag script skips `preventDefault` on a macOS double click, because it defers maximize to mouseup, and nothing then suppressed the browser's own word selection — double clicking the header highlighted the tab labels. The search field keeps `select-text`, since `select-none` on its row would otherwise reach into the input.

Also dropped the title bar's own double-click-to-maximize handler. Tauri's drag script already toggles maximize on a double click inside a drag region, so the local handler was a second toggle on the same gesture, and it keyed off the exact target, which no longer matches the area that now drags.

`npm run check:web` and `npm run build` pass. The drag itself needs a real window, so it wants a hands-on check on macOS, and on Windows/Linux for the double-click path.

### DIFF
--- a/src/chrome/ProjectRail.tsx
+++ b/src/chrome/ProjectRail.tsx
@@ -350,7 +350,7 @@ export function ProjectRail({
       className="sidebar-glass relative flex shrink-0 flex-col border-r border-content/10"
     >
       <div
-        className="flex h-10 shrink-0 items-center pr-1.5"
+        className="flex h-10 shrink-0 select-none items-center pr-1.5"
         data-tauri-drag-region="deep"
       >
         {IS_MAC ? <div className="w-[78px] shrink-0" /> : null}
--- a/src/chrome/ProjectRail.tsx
+++ b/src/chrome/ProjectRail.tsx
@@ -351,12 +351,10 @@ export function ProjectRail({
     >
       <div
         className="flex h-10 shrink-0 items-center pr-1.5"
-        data-tauri-drag-region
+        data-tauri-drag-region="deep"
       >
-        {IS_MAC ? (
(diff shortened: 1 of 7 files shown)


hardbeat920:
@emircan-sahin this is great. Have been meaning to do this myself as it was annoying. Thank you 🙏

PR #41, opened 2026-09-01
feat(linux): ship deb and AppImage releases

MisterWanted:
## What changed

- adds a shared Debian/Ubuntu Tauri dependency setup used locally and by CI
- adds explicit `.deb` and AppImage builds for tagged releases
- builds Linux first, transfers the packages as a workflow artifact, and only creates the public GitHub Release after both Linux and signed/notarized macOS artifacts are ready
- keeps macOS-only resources out of Linux bundles and sets Linux desktop category metadata
- suppresses the raw missing-updater-endpoint error; automatic checks stay quiet while manual checks point users to GitHub Releases

## Why

MonoCode already has Linux runtime support and Linux CI, but installable Linux builds required manual setup and tagged releases only shipped macOS. Source/Linux builds also intentionally have no updater endpoint configured.

The release dependency keeps publication atomic from GitHub's perspective: a failed Linux package build leaves no partial public release, so the workflow can be retried safely.

## UI

No layout change. A manual update check on a build without an updater feed now shows a short GitHub Releases message instead of the raw Tauri configuration error.

## Checklist

- [x] I ran `npm run check`

Additional verification:
- `npm run check`: 80 test files / 873 web tests; 147 Rust tests; TypeScript, fmt, clippy and doc tests green
- `actionlint .github/workflows/release.yml` passes cleanly
- built and inspected `.deb` and AppImage bundles; neither ships macOS `Assets.car`
- verified `Categories=Development;` in both Linux desktop entries
- package smoke test installed and launched successfully on Ubuntu amd64 / GNOME Wayland

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,9 +16,7 @@ jobs:
 
       - name: Install Linux system dependencies
         if: runner.os == 'Linux'
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y libwebkit2gtk-4.1-dev libgtk-3-dev libsoup-3.0-dev libjavascriptcoregtk-4.1-dev
+        run: bash scripts/install-linux-deps-debian.sh
 
       - uses: actions/setup-node@v4
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,7 @@ permissions:
 
 jobs:
   release:
+    needs: linux
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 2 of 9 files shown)


hardbeat920:
Thanks @MisterWanted, packaging looks good.

One thing before I merge. The auto-check on mount already swallows errors into idle, so this change only affects the manual click and it means "Check for updates" goes to "Checking…" and back with nothing shown. 

Agreed the raw error dialog was bad, but silence on an explicit click is worse. 
Can you keep the suppression and show a short message when manual is true, pointing to the GitHub releases?

MisterWanted:
Done — manual checks without updater endpoints now show a short GitHub Releases message, while the automatic mount check stays quiet. Real updater/network failures still follow the existing error path.

Added separate coverage for automatic vs manual missing-endpoint behavior. `npm run check` is green on this exact tree: 80 test files / 873 web tests and 147 Rust tests, plus fmt/clippy/doc tests. Pushed as `f585c60`.

hardbeat920:
This is great! 

One thing on the release job, linux runs after `gh release create`, so `/releases/latest` goes public with only the macOS files, and a failed Linux build cannot be retried because create already succeeded. 

Rest is fine.

MisterWanted:
Fixed the release ordering as requested. Linux now builds first and uploads its `.deb` + AppImage as a workflow artifact; the macOS release job has `needs: linux`, downloads those packages, and creates one GitHub Release containing both macOS and Linux assets. So a Linux failure leaves no public release and a rerun stays safe. I also removed the release workflow’s existing `ls | head` shellcheck warnings while touching that path. `actionlint` is clean and `npm run check` is green on the exact tree (873 web / 147 Rust). Pushed as `41fcb29`.

hardbeat920:
thank you @MisterWanted merged :)

PR #43, opened 2026-09-01
fix(skills): discover installed Claude plugin skills

### LINKED ISSUE #3: Installed plugins and skills are not displayed in MonoCode
## Description

Installed plugins and their skills are not appearing in MonoCode.

I have confirmed this with:

- Marketplace: `lucasAguiar11/agent-skills`
- Plugin: `workflow-kit`

The same plugin appears and works in Codex, Claude Code, and other environments. This may affect other installed plugins as well.

## Steps to reproduce

1. Install a plugin.
2. Open MonoCode.
3. Check the plugins or skills list.
4. The installed plugin and its skills are not displayed.

## Expected behavior

Installed plugins and their skills should appear in MonoCode.

## Actual behavior

The plugin and its skills are missing from the interface.

## Environment

- MonoCode version:
- macOS version:
- Mac model:
- Plugins tested: `workflow-kit`
- Screenshot or logs:

Repository: https://github.com/lucasAguiar11/agent-skills

MisterWanted:
## What changed

- discovers Claude plugin skills from registered plugin `installPath` entries in `~/.claude/plugins/installed_plugins.json`
- keeps plugin skills namespaced (for example `workflow-kit:quick-plan`) so they do not collide with normal project/user skills
- honors Claude `enabledPlugins` precedence: managed → project local → shared project → user
- supports file-based managed settings plus ordered `managed-settings.d/*.json` overrides
- respects project/local plugin scope for the configured project and its subdirectories
- ignores disabled plugins and stale plugin cache versions that are no longer registered
- fails closed when project-path canonicalization cannot be verified

## Why

Installed Claude plugins can provide skills under `<plugin>/skills/<skill>/SKILL.md`, but MonoCode only scanned the normal `.claude/skills` directories. That made installed plugin skills such as `workflow-kit` invisible in the MonoCode skill picker even though Claude Code could use them. Discovery also needs to follow Claude's enablement settings so a disabled plugin does not leak skills into MonoCode.

Fixes #3.

## UI

No layout change. Enabled plugin-provided skills now appear in the existing skill picker under their Claude plugin namespace.

## Checklist

- [x] I ran `npm run check`

Additional verification:
- rebased onto current upstream `main` / MonoCode 0.1.25 before the final check
- `npm run check`: 89 test files / 958 web tests; 163 Rust tests; fmt, clippy and doc tests green
- plugin-specific Rust coverage: installed discovery, namespace collision avoidance, user/project/local precedence, explicit `false` hiding skills, managed drop-in ordering, project scoping from nested cwd, stale cache filtering, and fail-closed path checks
- verified the reported `workflow-kit` repository uses the expected `plugins/workflow-kit/skills/<skill>/SKILL.md` layout

### DIFF
--- a/src-tauri/src/skills.rs
+++ b/src-tauri/src/skills.rs
@@ -72,11 +72,199 @@ pub(crate) fn list_skills_from(project: &Path, home: Option<&Path>) -> Vec<Disco
     if let Some(home) = home {
         add_root(home.join(".pi/agent/skills"), "user", "pi");
         add_root(home.join(".omp/agent/skills"), "user", "omp");
+        for (root, scope, namespace) in claude_plugin_skill_roots(home, project) {
+            add_namespaced_root(&mut by_name, root, scope, "claude", &namespace);
+        }
     }
 
     let mut out: Vec<DiscoveredSkill> = by_name.into_values().collect();
     out.sort_by(|a, b| a.name.cmp(&b.name));
     out
+}
+
+fn add_namespaced_root(
+    by_name: &mut HashMap<String, DiscoveredSkill>,
+    root: PathBuf,
+    scope: &str,
+    source: &str,
+    namespace: &str,
+) {
+    if by_name.len() >= MAX_SKILLS {
+        return;
+    }
+    for mut skill in scan_root(&root, scope, source) {
+        if by_name.len() >= MAX_SKILLS {
+            break;
+        }
+        skill.name = format!("{namespace}:{}", skill.name);
+        by_name.entry(skill.name.clone()).or_insert(skill);
+    }
+}
+
+fn claude_plugin_skill_roots(home: &Path, project: &Path) -> Vec<(PathBuf, &'static str, String)> {
+    let registry = home.join(".claude/plugins/installed_plugins.json");
+    let Ok(raw) = std::fs::read_to_string(registry) else {
+        return Vec::new();
+    };
+    let Ok(value) = serde_json::from_str::<serde_json::Value>(&raw) else {
+        return Vec::new();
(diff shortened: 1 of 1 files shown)


hardbeat920:
Thanks @MisterWanted the registry-based discovery and plugin:skill namespacing look right!

Before merge, plugin skills should honor Claude’s enablement layer, not only `installed_plugins.json`

Claude treats `enabledPlugins` in managed / local / project / user settings as the actual on/off switch. 
`false` means the plugin stays on disk but does not load, and its skills go away with it. 

Please skip disabled plugins (same precedence Claude uses), and add a test that "workflow-kit@community": false hides those skills.

Smaller follow-up in `path_is_within`:
if `canonicalize` fails, return `false` instead of a lexical `starts_with`. The nested-cwd test still works when both paths exist; the fallback can match too broadly

No other concerns from the diff. Happy to approve once enablement is in.

MisterWanted:
Addressed both review points and rebased onto current `main` (0.1.25). Plugin discovery now resolves `enabledPlugins` per plugin with Claude precedence: managed → project local → shared project → user; an explicit `"workflow-kit@community": false` hides its skills. File-based managed settings include ordered `managed-settings.d/*.json` overrides. `path_is_within` now returns false if either canonicalization fails. Added coverage for disabled plugins, precedence, managed drop-ins, and fail-closed paths. Full `npm run check` on the rebased exact tree is green: 89 test files / 958 web tests and 163 Rust tests, plus fmt/clippy/doc tests. Pushed as `231dfac`.

hardbeat920:
this is great @MisterWanted 
Thank you

PR #44, opened 2026-09-01
fix(shortcuts): stop focused agent turn on Escape

MisterWanted:
## What changed

- makes plain Escape stop the active focused in-flight agent turn
- reuses the existing `onStop` path, including provider cancellation and local session cleanup
- never treats Escape as PTY cancellation; terminal panes keep normal terminal input
- yields to UI surfaces that already own Escape, including modals, pickers, search/editor controls, and drag/reorder cancellation
- only stops an actually busy agent pane; diff/editor/project-terminal focus, stale active tabs, held-key repeats, modifiers, and IME composition are ignored
- re-checks the active/focused busy session after the key event dispatch before stopping, so focus changes or a later Escape consumer win

## Why

Issue #4 was clarified to mean stopping an in-flight agent turn, not interrupting a shell command. PTY commands continue to use Ctrl+C; Escape becomes the quick stop shortcut for the active agent turn.

Fixes #4.

## UI

No layout or visual change, so there is no meaningful before/after screenshot.

## Verification

- exact head: `8add56753b527429ad9897f4b8a54620f442ad6d`
- exact upstream base: `2daf1a19649ed9eff87936fe60feb7ed4cad19d8` (MonoCode 0.1.27)
- one commit, three changed files; no unrelated repair code
- stable patch-id is unchanged across the final upstream rebase
- local `npm run check:web`: 94/94 test files, 1020/1020 tests
- targeted shortcut tests: 14/14 green
- `git diff --check`: clean
- exact-head GitHub CI: Ubuntu green and macOS green, including `npm test`, `tsc --noEmit`, `cargo fmt --check`, Clippy with `-D warnings`, `cargo check`, and `cargo test`
- negative-space review confirmed existing Escape consumers call `preventDefault`, so their behavior takes precedence over agent cancellation

## Checklist

- [x] Exact-head CI is green on all configured platforms
- [x] This PR is small and focused
- [x] I did not mix unrelated changes

PR #45, opened 2026-09-01
fix(chrome): restore development label helpers

MisterWanted:
## What changed

- exports `DevModeLabel` for the new Sidebar import
- adds the missing `DevModeSlot` used by ProjectRail and Sidebar
- keeps the development badge in the explicit chrome slots instead of rendering it twice inside `TabVisitNav`

## Why

Current `main` at `0c63d7d` does not typecheck: `ProjectRail.tsx` and `Sidebar.tsx` import helpers that `TitleBar.tsx` does not export/define. This also causes PR CI based on the current merge ref to fail before reaching Rust checks.

## Verification

- `npm run check:web`: 93/93 test files, 985/985 tests
- `npx tsc --noEmit`: green
- `git diff --check`: green

This is intentionally separate from PR #44 so that shortcut PR stays scope-clean.

### DIFF
--- a/src/chrome/TitleBar.tsx
+++ b/src/chrome/TitleBar.tsx
@@ -752,7 +752,7 @@ export function IconButton({
   );
 }
 
-function DevModeLabel() {
+export function DevModeLabel() {
   if (!import.meta.env.DEV) return null;
   return (
     <span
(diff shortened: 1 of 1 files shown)


MisterWanted:
Superseded by upstream main commit `f013713` which fixes the missing development-label exports/slot directly. Closing this duplicate to keep review noise low.

PR #47, opened 2026-09-01
feat(shortcuts): navigate sessions and projects

### LINKED ISSUE #21: There's an ordering bug when switching sessions with keyboard shortcuts
### MonoCode version

latest

### macOS version

26.4.1

### Provider CLI

Multiple / not sure

### What happened

While working with multiple sessions, I noticed something when using CMD+1, CMD+2, CMD+3. When I press CMD+3, it switches to the 2nd one in the list. Basically, the order shown in the UI and the order the shortcuts actually follow are completely out of sync.

### Steps to reproduce

While working with multiple sessions, I noticed something when using CMD+1, CMD+2, CMD+3. When I press CMD+3, it switches to the 2nd one in the list.

MisterWanted:
Fixes #21

Went with the spatial version from the issue:
- cmd+shift+↑/↓ cycles sessions in the sidebar order
- cmd+shift+←/→ cycles projects in the rail order

Kept cmd+1…9 as tabs. The new shortcuts stay out of text fields, editors, terminals and overlays.

Session cycling also follows filters/folders/collapsed state instead of just whatever happens to be mounted on screen.

Tested with `npm run check:web` and `npm run build`.


hardbeat920:
This looks great can you rebase before merge. Thanks again 🙏

truongio:
Some feedback, would it be possible to make these keys rebindable? I use Hyper key a lot (CMD + Option + Ctrl) and stay on home row a lot so moving finger to arrows would be slow. I don't know how common it is but my "arrow" keys are Hyper + h/j/k/l and that wouldn't work with the combo above since Hyper interferes with that.

hardbeat920:
Good point @truongio! Configurable keybindings make sense, but shortcuts are currently hard-coded and the Keybindings settings page is read-only, so doing this properly would be a broader feature. 

I’d prefer to land this PR with the proposed defaults and track customizable keybindings separately. 

Thanks for flagging it.

PR #48, opened 2026-09-02
Add browser remote access

namphamdev:
## Summary

- Add a token-authenticated HTTP/WebSocket remote server in the Tauri backend so the desktop app can be controlled over the network
- Introduce a web bridge layer (`src/bridge/*`) that lets the existing UI run in a browser when connected remotely
- Add Settings UI to enable/disable remote access, manage the access token, and copy connection URLs

## Test plan

- [ ] Build and run the desktop app locally
- [ ] Enable remote access in Settings and verify a token + URL are shown
- [ ] Open the remote URL in a browser with the token and confirm the UI loads
- [ ] Exercise core flows over remote (PTY, invoke commands, events) and confirm blocked window-only commands stay unavailable
- [ ] Disable remote access and confirm the server stops and browser sessions disconnect

Made with [Cursor](https://cursor.com)


hardbeat920:
thanks for putting this together @namphamdev 

Remote access is something I want in MonoCode, and this is a useful spike.

I’m not going to merge it as-is though. Getting this right is a lot bigger than serving the current desktop UI in a browser. The desktop shell is built around Tauri (windows, dialogs, native chrome, the full IDE layout), and remoting that over HTTP/WebSocket turns every invoke into a network API. That’s hard to secure and harder to maintain, and it still doesn’t work well on a phone.

I am open to brainstorm about this first in an issue/discussion! I am thinking a small host protocol, and a separate companion UI meant for mobile (running agents, follow-up, inbox) rather than a squeezed copy of the desktop app.

Really appreciate the work but i will need to close this PR. Let's open a discussion on this and move from there

PR #49, opened 2026-09-02
feat(inbox): show item status in the list and fix the status filter

emircan-sahin:
Two related inbox changes: the list never showed an item's status, and the status filter briefly showed the right list before snapping back.

## Status in the list

Every row drew the same neutral kind icon, so open, draft, merged and closed items looked identical until you opened one. The icon now carries the status in both its glyph and its colour:

| Status | Icon | Colour |
| --- | --- | --- |
| Open | `GitPullRequest` / `CircleDot` | green |
| Draft | `GitPullRequestDraft` | grey |
| Merged | `GitMerge` | purple |
| Closed | `GitPullRequestClosed` / `CircleX` | red |

The glyph changes with the status, so the meaning survives colour blindness, and the card's `aria-label` now names the status for screen readers. The detail header reuses the same map, which also gives closed items their own colour instead of the grey they used to share with drafts.

Inbox list with status colours

## Filter no longer snaps back to open

`filterInboxByStatus` reads an empty status filter as "no restriction", but `inboxFetchState` asked the providers for open items only unless Closed or Merged was checked. The two layers disagreed, so unchecking Merged rendered the previous `all` response unfiltered and then, a moment later, replaced it with an open-only refetch.

The fetch now widens with the filter:

- nothing checked -> `all`
- only Open and/or Draft checked -> `open`
- Closed or Merged checked -> `all`

Because an unfiltered fetch competes with closed history for the same page, `all` now asks for 100 rows instead of 40 (`INBOX_ALL_LIMIT`, passed to both GitHub and Linear), so open work is not crowded out of the inbox.

One consequence worth flagging: the inbox unseen badge runs the same query, so with no status filter it can now light up for merged and closed activity too. That keeps the badge consistent with what the list shows, but happy to pin it to open items if you would rather it stayed quiet.

## Testing

`npm run check:web` - 1047 tests pass, `tsc --noEmit` clean.

### DIFF
--- a/src/chrome/icons.tsx
+++ b/src/chrome/icons.tsx
@@ -19,6 +19,7 @@ import BrainIcon from "@hugeicons/core-free-icons/BrainIcon";
 import Cancel01Icon from "@hugeicons/core-free-icons/Cancel01Icon";
 import CaseSensitiveIcon from "@hugeicons/core-free-icons/CaseSensitiveIcon";
 import CircleArrowDown01Icon from "@hugeicons/core-free-icons/CircleArrowDown01Icon";
+import CancelCircleIcon from "@hugeicons/core-free-icons/CancelCircleIcon";
 import CircleDashedIcon from "@hugeicons/core-free-icons/CircleDashedIcon";
 import CircleDotIcon from "@hugeicons/core-free-icons/CircleDotIcon";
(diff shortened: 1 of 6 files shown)


hardbeat920:
This is a great little UX improvement. I love it!

PR #51, opened 2026-09-02
Add an opt-in Reply pace setting for smoother streamed replies

emircan-sahin:
## What changed

Settings > General > **Reply pace** — an opt-in setting that streams a reply smoothly instead of in jumps of five or ten lines. It defaults to **Off**, which is exactly what the transcript does today.

## Why

The jumps are not a rendering problem, they are the shape of the input. Measured against the Claude CLI in the mode MonoCode spawns it (`--output-format stream-json --input-format stream-json --include-partial-messages`, sonnet):

```
text_delta events: 6      total chars: 478
average gap: 919ms        p50: 553ms
average per event: 79.7 characters
effective updates/sec: 2.4
```

At raw stdout level the same: 355–643 byte writes with 500–950ms of silence between them. Not pipe buffering (the chunks are far under the buffer size) — the CLI genuinely writes in bursts. So the transcript receives ~80 characters at a time, about twice a second, and applying each delta whole drops a paragraph in at once.

The app side was already fine: Rust emits per line, and the frontend already batched deltas into one animation frame. There was simply nothing finer to render.

The setting cuts each burst at a character budget and applies the rest over the following frames. The budget is wall-clock, not per-frame — a per-frame budget assumes 60Hz, and the frames drop under exactly this load, which stretched the reveal past the end of the reply. Simulated against the measured arrival pattern, a 3000-character answer on Smooth finished 0.6s after the agent at 60Hz but 2.2s at 20Hz and 4.2s at 12Hz; against elapsed time it is 0.6s at every frame rate.

Against that same arrival pattern the three settings measure:

| | frames painting text | longest silence | slice | updates/sec |
|---|---|---|---|---|
| Off | 5% | 420 ms | 80 chars | 2.4 |
| Balanced | 68% | 133 ms | 4.7 ± 2.4 chars | 41 |
| Smooth | 100% | 0 ms | 3.2 ± 1.5 chars | 60 |

The extra cost is one React/markdown pass per update: `parseMarkdownIntoBlocks` over the full text (0.20 ms at 5k characters, 0.76 ms at 20k) plus the trailing block's render (0.48 ms, and flat in answer length, because Streamdown memoises completed blocks). At 60 updates/sec that is roughly 40 ms of work per second of streaming — a few percent of one core, and in practice small enough that it does not clearly separate from the app's baseline in Activity Monitor. It is still real work that only pays off if you want the smoother stream, so the setting is off unless you turn it on.

There is no "slow" option because there cannot be one: bounding how far the reveal falls behind pins its rate to the model's. A profile tuned to trail measured 175 chars/s against Smooth's 188 — indistinguishable. Genuinely slower means trailing a long answer by half a minute.

## UI

Left is the released app, right is this branch on **Smooth**. Same prompt, same model.

New row in Settings > General:

### Off is not quite a no-op

With Off the text merges and renders exactly as it does on main, but four ordering fixes in the queue apply regardless of the setting. All four have a window of at most one animation frame today, so they are latent rather than visible:

- The per-frame commit is now functional and touches only the sessions that changed. It was a value-based `setSessions(next)` built from a ref snapshot, which could overwrite a `stopStreaming` queued from a promise continuation for a different session, leaving that session stuck busy with a streaming block.
- A steer, a harness switch, and a new turn now drain the session's queue before writing the user block. Previously a delta queued in the same frame landed *under* the new prompt.
- Quit drains the queue before the snapshot is written, and the unload path drains instead of cancelling, so a frame's worth of text can no longer be missing from the saved session.
- Turn end and Stop drain only their own session instead of every session.

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Replies stream in smoothly instead of arriving in jumps of five or ten lines. The agent CLIs hand over a paragraph at a time — Claude's sends about 80 characters 2.4 times a second — and the transcript now spreads each burst across frames as it arrives. Settings > General > Reply pace picks how finely: Smooth (the default), Balanced for a third less work per second, or Off for the old behaviour.
+- Settings > General > Reply pace can stream replies smoothly instead of in jumps of five or ten lines. The agent CLIs hand over a paragraph at a time — Claude's sends about 80 characters 2.4 times a second — so Smooth spreads each burst across frames as it arrives, and Balanced does the same in coarser steps for about a third less work. It defaults to Off, which is the existing behaviour.
 
 ## [0.1.29] - 2026-09-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Replies stream in smoothly instead of arriving in jumps of five or ten lines. The agent CLIs hand over a paragraph at a time — Claude's sends about 80 characters 2.4 times a second — and the transcript now spreads each burst across frames as it arrives. Settings > General > Reply pace picks how finely: Smooth (the default), Balanced for a third less work per second, or Off for the old behaviour.
+
 ## [0.1.29] - 2026-09-02
 
 ### Added
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -123,6 +123,7 @@ import {
   startHarnessBridge,
   stopStreaming,
   pickTextHarness,
+  DRIP_FRAME_MS,
   STREAM_PACE_DEFAULT,
   type ApprovalDecision,
   type HarnessEvent,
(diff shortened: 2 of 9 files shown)


emircan-sahin:
I set smooth mode to off by default since some developers will want lower CPU usage.

If someone's a developer reviewing code and doesn't want eye strain, they can turn smooth mode on. On my end, running AI in 2 tabs at the same time only uses about 3.5% CPU, and it's roughly the same with it off too.

https://github.com/user-attachments/assets/a36bab60-12e8-4377-9878-a7de9e46b2ae

hardbeat920:
This is great! However, we use streamdown... I wonder if we can use the animated streaming options that come with it already

https://streamdown.ai/docs/animation

emircan-sahin:
> This is great! However, we use streamdown... I wonder if we can use the animated streaming options that come with it already
> 
> https://streamdown.ai/docs/animation

I tried the library you used and added the animation, but it introduces way too much lag, and it becomes a lot more noticeable in lists. For example, while index number 16 is already displayed, the animation is still stuck rendering index 10 because of the delay.

On top of that, it adds extra strain in smooth mode, making it feel like the FPS is dropping, so it stutters as it progresses, adding even more lag. Because of that, I ended up removing the animation again.

If you like it, feel free to merge it, if not, you can turn it off. It's entirely a matter of personal preference.

hardbeat920:
@emircan-sahin thanks for this. 
After reconsidering, I don’t think I want to add a Reply pace setting and the associated streaming complexity right now.

The relevant event-handling code on main has also changed considerably, so I’m going to close this PR.

We can reconsider in the future 🙏

PR #52, opened 2026-09-02
Show, stop and message Claude Code subagents from a panel

emircan-sahin:
Claude Code runs subagents, but MonoCode only ever showed them as a collapsed "Running 2 subagents" line: no way to see what each one was doing, read what it wrote, or stop one. This adds a subagents panel for Claude sessions, on top of a harness-neutral run model so other harnesses can plug in later.

**Scope: Claude Code only.** Codex, Cursor, OpenCode and the rest keep their current behaviour; the panel never appears for them because nothing emits run events yet. See "Other harnesses" below for what it takes.

## What the user gets

- **Strip above the composer** while subagents run: spinner, "N subagents running", the current activity of the first one, and Open. It disappears when nothing runs, so a stale count can no longer read as "still running".
- **Subagents panel** (right drawer): runs as a tree (nested spawns indented under their parent), each with activity, tokens and elapsed time; finished runs dimmed with done / stopped / failed. Selecting a run shows its brief, its own transcript (text, thinking, tool calls) and a reply box.
- **Stop / Stop all**, with the CLI's verdict read back. While a stop is in flight the button reads "Stopping…" and is disabled; a refusal is written into the transcript instead of vanishing.
- **Turn footer** shows how many subagents that turn spawned ("Worked for 21s · 19:21 · 3 subagents"); clicking it opens the panel on the first one.
- **Open** on the transcript row of any real subagent call.

## Protocol work, measured against CLI 2.1.258

The initialize request now declares `forwardSubagentText`, `agentProgressSummaries` and `perTaskStopAffordance`. Forwarded frames carry `parent_tool_use_id`, which is the only id a subagent's output ever names; a run is therefore keyed by the spawning Agent tool call, with the task id attached when `task_started` arrives.

Things the CLI does that shaped the code:

- `background_tasks_changed` arrives *before* `task_started` for the same task, so nothing is keyed from it; the earlier code minted a second row there, which is where the "6 rows for 3 agents" bug came from.
- `task_progress.description` is the agent's current activity ("Reading src/App.tsx"), not its name. It now feeds the detail line; the name stays what the spawn said.
- `stop_task` works on foreground and background agents, but does **not** stop a killed agent's own children. Stops cascade leaves-first.
- With `perTaskStopAffordance` declared, `interrupt` spares background agents, so the session Stop now sends `stop_task` for every live run before the interrupt.
- The SendMessage address in the Agent tool result is the task id, so a run only ever seen through forwarded frames still has a stop handle.

Two bugs in the previous Stop path were the cause of every "did not answer in time": after Stop the harness muted its whole stream, including `control_response`, and the app dropped every event for the old turn generation, including "run stopped". The mute now covers only the cancelled turn's own trailing output, and agent lifecycle events pass the generation guard, since subagents legitimately outlive the turn.

## Messaging a subagent

There is no host-to-subagent channel in the CLI's control protocol; the only address a subagent answers on is the model's SendMessage tool. The panel's reply box therefore sends the main agent a relay instruction ("forward this to subagent X, do not act on it"). It works, but it is indirect, and worth knowing before relying on it.

## Other harnesses

The run model (`AgentRun`, the `agent.*` harness events, the reducer, the panel, strip and footer) is harness-neutral. A harness joins by emitting `agent.started` / `agent.updated` / `agent.output` / `agent.tool` from its own stream and, optionally, implementing the two new adapter hooks `stopAgent` and `agentRelayPrompt`. Codex already exposes `subAgentActivity` items with an agent path, so it would be a mapping job in `codexProtocol.ts`, not new UI.

Subagent transcripts live in memory for the session; they are not persisted yet.

## Testing

`npm run check:web` - 1084 tests pass (26 new across the run reducer, protocol parsers and a mocked live CLI: identity across `background_tasks_changed`/`task_started`, cascade stops, refused stops, stops after cancel, stale-run settling, one row per agent). `tsc --noEmit` clean, `npm run build` OK. The lifecycle claims above were checked with probe scripts against the real CLI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Claude Code subagents get their own panel. A strip above the composer shows how many are running and what the first one is doing, and opens a right-hand drawer that lists every run as a tree (nested spawns under their parent) with activity, tokens and elapsed time. Selecting a run shows its brief and its own transcript — text, thinking and tool calls — plus a reply box that relays a message through the main agent. Stop one run or Stop all; the button reads "Stopping…" until Claude confirms, and a refusal is written to the transcript instead of vanishing. The turn footer counts the subagents each turn spawned and opens the panel on click. Claude only; other harnesses are unchanged. Subagent transcripts are kept for the session, not persisted.
+
 - Settings → General → Diff view: Editor or Unified. Unified stacks every working-tree change in one **Changes** tab — GitHub-style review, editor syntax colours, sticky file headers and line numbers, and a single horizontal scroll that stops at the end of the line. Editor keeps the previous per-file working-tree tabs.
+
+### Fixed
+
+- Claude subagents no longer appear as several transcript rows. Progress and background-task updates named the agent's current activity ("Running find …"), and each one was mistaken for a new agent.
 
 ## [0.1.29] - 2026-09-02
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -121,12 +121,15 @@ import {
   sendHarnessTurn,
   steerHarnessTurn,
   startHarnessBridge,
+  stopHarnessAgent,
+  harnessAgentRelayPrompt,
   stopStreaming,
   pickTextHarness,
   type ApprovalDecision,
   type HarnessEvent,
   type UserQuestionReply,
 } from "./lib/harness";
+import { isAgentEvent } from "./lib/harness/agentRuns";
 import {
   appendPreparingHandoff,
   buildDeterministicHandoff,
@@ -200,6 +203,7 @@ import {
   sessionDisplayTitle,
   sessionWorkCwd,
   titleFromPrompt,
+  type AgentRun,
   type Attachment,
   type Block,
   type HarnessId,
@@ -3227,7 +3231,13 @@ export default function App({
               : prompt,
             attachments: prepared,
             onEvent: (event) => {
-              if (turnGen.current.get(sessionId) !== gen) return;
+              if (turnGen.current.get(sessionId) !== gen) {
+                // Stop bumps the generation before the harness has answered,
+                // and subagents can outlive the turn. Their lifecycle is not
+                // turn output: drop it and Stop leaves runs "running" forever.
+                if (isAgentEvent(event)) enqueueHarnessEvent(sessionId, event);
+                return;
+              }
               if (
                 wrap &&
                 (event.type === "session.started" ||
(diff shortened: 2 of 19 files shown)


emircan-sahin:
I tried to build a clean system while also keeping the UI/UX solid. I also structured the architecture with future OpenAI model support in mind, so if you ever want to add that, all you'd need to do is review the architecture and slot it into the existing conditions.

Ekran Resmi 2026-09-02 19 25 24

Ekran Resmi 2026-09-02 19 25 34

hardbeat920:
This is an interesting feature and definitely worth exploring. That said, I think it would be worth discussing the UX implications a bit more before implementing it, so we don’t accidentally paint ourselves into a corner later on.

In particular, I’m wondering about the case where I want to open a panel next to the main one. 
For example, to view files, inspect changes, or start a second chat. Would that panel end up being hidden behind the subagent panel?

I like the idea overall, but I think features like this are worth discussing before implementation, as they could have broader implications for the tool’s UX and potentially affect how we approach panels more generally.

emircan-sahin:
> This is an interesting feature and definitely worth exploring. That said, I think it would be worth discussing the UX implications a bit more before implementing it, so we don’t accidentally paint ourselves into a corner later on.
> 
> In particular, I’m wondering about the case where I want to open a panel next to the main one. For example, to view files, inspect changes, or start a second chat. Would that panel end up being hidden behind the subagent panel?
> 
> I like the idea overall, but I think features like this are worth discussing before implementation, as they could have broader implications for the tool’s UX and potentially affect how we approach panels more generally.

Yeah, we genuinely need to talk this through, either keep working on this branch or you should accept the PR, and depending on how it goes, we can always revert or refine it later. When you open multiple sessions, each session's subagent opens in its own panel, so nothing gets broken there. But things like the diff view still matter.

Maybe when a subagent gets clicked, we should shrink the diff panel, or if the diff panel gets expanded, the subagent panel could collapse, something along those lines.

emircan-sahin:
Subagent panel with multiple sessions.

Ekran Resmi 2026-09-02 20 18 18

emircan-sahin:
Subagents with diff panel.

Ekran Resmi 2026-09-02 20 26 47

hardbeat920:
I would be much more inclined to show the subagent work in a more minimal way rather than a full on panel of its own.

I am not sure i want to inspect every single detail of the subagent. This is why i decided to make the tool calls more minimal compared to how they were initially.

How i think about it to have a ticker showing the steps the subagents are taking right where the tool calls are, by default it can just tick but you can expand if you want to get more details.

From the screenshots that you sent i can see that is becoming a bit busy when you have more things going on... also if you reduce the transparency of the app how does it look? Is the subagent panel transparent with blur? and if so how readable it is while sitting on top of the session panel

emircan-sahin:
> I would be much more inclined to show the subagent work in a more minimal way rather than a full on panel of its own.
> 
> I am not sure i want to inspect every single detail of the subagent. This is why i decided to make the tool calls more minimal compared to how they were initially.
> 
> How i think about it to have a ticker showing the steps the subagents are taking right where the tool calls are, by default it can just tick but you can expand if you want to get more details.
> 
> From the screenshots that you sent i can see that is becoming a bit busy when you have more things going on... also if you reduce the transparency of the app how does it look? Is the subagent panel transparent with blur? and if so how readable it is while sitting on top of the session panel

There are actually two ways to go about this.

1. Accept what's already there and add a setting for minimizing/maximizing the subagent view.
2. Close this one out, build just the minimized version, and while doing that, look at how this PR handles the SDK/API and what considerations it takes into account.

hardbeat920:
Thank @emircan-sahin the protocol notes are great. I'm going to close this rather than merge. 

The direction I want is a small ticker on the existing tool-call row, expandable for more detail. I'll implement it and credit you for the protocol work. 

No need to keep iterating on this branch 🙏

PR #56, opened 2026-09-03
feat: add configurable queued follow-ups

tcmarkfeld:
## Summary

- Add Queue and Steer follow-up behavior setting
- Keep Steer as default behavior
- Send queued messages after active turn completes
- Pause queue when active turn is stopped
- Resume interrupted turn before queued messages continue
- Add edit, steer, and remove actions for queued messages
- Match queue styling with stacked composer controls
- Standardize segmented control widths

image
image
image

## Verification

- Full test suite passed: 1,083 tests
- TypeScript check passed
- Production build passed


hardbeat920:
Looks great @tcmarkfeld  no issues with the Steer path, and Queue as an opt-in setting makes sense.

Two things on the queue dispatcher before this lands:

1. The head item gets removed from the queue before `onSubmit` actually runs. If submit no-ops (e.g. a preparing handoff) or another send lands in that same tick, the follow-up can disappear or get steered into the wrong turn.

2. `editingQueuedMessageId` skips dispatch for the whole session, not just the row being edited. If the turn finishes while you’re editing item 2 of 3, nothing in the queue goes out until you save or cancel. Only hold dispatch when the head item is the one being edited.

Happy to re-review after that.

tcmarkfeld:
@hardbeat920 updated

hardbeat920:
thanks @tcmarkfeld this look great. I love the feature as well

PR #62, opened 2026-09-03
Add opt-in desktop notifications for finished and waiting sessions

emircan-sahin:
## What changed

A **Notifications** toggle in Settings, off by default. When on, a system notification appears when a turn finishes or an agent waits on an approval or question in a session that is not on screen, whether MonoCode is in the background or another session is open. Clicking the banner or its **Show** button focuses the window and opens that session.

- macOS uses `UNUserNotificationCenter` directly (already linked for the Dock badge): real permission state, a `Show` action category, and a delegate that presents banners while the app is frontmost and routes clicks back to the session. When alerts are blocked, the Settings row shows "Permission needed" with an **Open System Settings** button that deep-links to MonoCode's notification page.
- Linux uses `notify-rust` over the freedesktop bus; no permission model there. Other platforms report `unsupported` and the toggle says so.
- The **Sounds** setting decides whether the notification carries a sound. When a banner fires, the in-app turn-finished cue is skipped so nothing chimes twice.
- Notification layout: title `MonoCode`, subtitle is the session title, body is the first paragraph of the reply (or the pending approval / question).

Also fixes `tauri dev` on macOS: the dev bundle is now ad-hoc signed with `com.monocode.desktop`. The linker's signature carried a `monocode-<hash>` identifier, and `UNUserNotificationCenter` refuses authorization without prompting when that disagrees with `CFBundleIdentifier`. The binary is copied instead of hard-linked because re-signing rewrites the file the running process was linked to.

## Why

Sessions often finish or stall while you are in another app or another tab. Sounds already cover the first case; a clickable banner covers the second and gets you back to the right session in one click.

## UI

New Settings row under Sounds: "Notifications" toggle, with "Permission needed · Open System Settings" beside it when macOS has alerts blocked.

## Checklist

- [x] I ran `npm run check` (the two `git_stash_*` Rust tests fail on `main` for me too, unrelated)


emircan-sahin:
I think everyone's going to love this feature.

I was using cmux purely for the notifications, now MonoCode will have that too 🙈

hardbeat920:
Thanks for the PR @emircan-sahin. Before merging, could you address these issues:
1. macOS requests badge-only permission at startup, consuming the one-time prompt before Notifications is enabled.
2. Linux waits for a "default" click action but never registers one.
3. Linux notification bodies should escape/strip markup because they contain untrusted agent output.
4. Only suppress the completion sound after notification dispatch succeeds.

emircan-sahin:
Thanks for the review, all four addressed in 93604cf:

1. **macOS startup prompt**: `request_badge_authorization` now checks `getNotificationSettings` first and only re-requests the badge option when the user has already answered. While the status is still undecided it does nothing, so the one-time system dialog is reserved for the Notifications toggle (which asks for alert + sound + badge together). The Dock badge starts working once that prompt has been answered.
2. **Linux click**: the notification now registers `.action("default", "Show")`, so servers actually report the click that `wait_for_action` was waiting on.
3. **Linux markup**: bodies go through an `escape_markup` helper (`&`, `<`, `>`) before dispatch, with a unit test.
4. **Sound suppression**: `notifySession` is now async and resolves `true` only after the `show_notification` invoke succeeds. The turn-finished cue plays whenever it resolves `false`, including a rejected dispatch.

hardbeat920:
Thanks @emircan-sahin, the permission flow and Linux issues look fixed. 

One remaining issue: macOS show_notification returns success before its authorization and scheduling callbacks complete, so the frontend can still suppress the completion cue when no notification was delivered. 

Could the command await the actual scheduling result before returning success?

emircan-sahin:
Fixed in 989b97a. `show_notification` is now an async command that resolves only after macOS reports the outcome:

1. It first reads `getNotificationSettings`. Anything other than authorized-with-alerts (denied, undetermined, or badge-only) returns an error immediately, so the frontend keeps its cue. This also drops the old `requestAuthorization` call in the show path, which means a finished turn can no longer pop the system prompt.
2. It then hands the request to the center and awaits `addNotificationRequest`'s completion handler, returning the `NSError` description when the center rejects it.
3. Both waits are bounded by a 5s timeout, so a handler that never runs falls back to an error instead of leaving the promise pending and the turn silent. The permission prompt itself still waits indefinitely, since that one is gated on the user answering the dialog.

Linux and the unsupported fallback became async to match; their behaviour is unchanged, as `notify_rust`'s `show()` was already propagating failures synchronously. Two unit tests cover the new wait plumbing (value delivered, and the give-up path).

PR #67, opened 2026-09-04
Refresh @ file mentions for new and spaced paths

### LINKED ISSUE #66: @ mentions miss newly pasted files and paths with spaces
### MonoCode version

0.1.32

### macOS version

26.4

### Provider CLI

Multiple / not sure

### What happened

I expected files I copy into the project (including a folder of images, and names with spaces) to show up in the composer `@` picker when I return to MonoCode and type `@`.

Instead:

1. Newly pasted files do not appear until something else reloads the project file list. `@` keeps using a stale in-memory index.
2. Files and folders whose names contain spaces never appear in `@` at all.

### Steps to reproduce

1. Open a project in MonoCode.
2. Copy a new file into the project directory from Finder, and copy a folder whose name has a space (for example `My Photos/cat.png`).
3. Switch back to MonoCode.
4. Type `@` in the composer and search for those names.
5. The new file is missing, and the spaced path is missing even after a reload.

elanchezhiyanr:
## What changed

The project file index now refreshes on window focus and directory-change notifications, and a refresh is no longer dropped if a scan is already in flight. The composer `@` picker subscribes to those updates.

Paths with spaces are included in `@`. The picker still shows the real name. Picking one inserts a single-word `@` token (spaces become hyphens). On send, the prompt maps that token to the real project-relative path. Paths that leave the workspace or would break the tokenizer are skipped.

## Why

`@` used a stale in-memory `git ls-files` cache. The explorer already reloaded on focus; the mention list did not. Files copied into the project from Finder therefore did not appear until something else forced a full reload.

`@` tokens are one whitespace-delimited word, so spaced names were excluded entirely. They now stay a normal `@label` in the composer, with the real path recovered from the mention index on send rather than decoded from user text.

## UI

Behavior only. Copy a file or a folder (including names with spaces) into the project, return to MonoCode, type `@` plus part of the name, pick it, and confirm it is referenced in the prompt.

## Checklist

- [x] I ran `npm run check`

Fixes #66

### DIFF
--- a/src/chrome/Composer.tsx
+++ b/src/chrome/Composer.tsx
@@ -40,6 +40,7 @@ import {
   loadProjectFiles,
   peekProjectFiles,
   recentOpenedFiles,
+  subscribeProjectFiles,
 } from "../lib/fileIndex";
 import {
   buildMentionIndex,
@@ -601,13 +602,21 @@ export function Composer({
 
   useEffect(() => {
     let cancelled = false;
+    const apply = (next: ProjectFile[]) => {
+      if (!cancelled) setFiles(next);
+    };
+    const cached = peekProjectFiles(cwd);
+    if (cached) apply(cached);
     void loadProjectFiles(cwd, mentionOpen)
-      .then((next) => {
-        if (!cancelled) setFiles(next);
-      })
+      .then(apply)
       .catch(() => undefined);
+    const unsub = subscribeProjectFiles(() => {
+      const next = peekProjectFiles(cwd);
+      if (next) apply(next);
+    });
     return () => {
       cancelled = true;
+      unsub();
     };
   }, [cwd, mentionOpen]);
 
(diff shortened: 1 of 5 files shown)


hardbeat920:
Looks great overall @elanchezhiyanr. Could you reject Unicode line separators and bidi control characters in isSafeProjectRelative() and add a regression test?

elanchezhiyanr:
Hi @hardbeat920 Thanks. Done in the latest commit.

`isSafeProjectRelative()` now rejects Unicode controls, bidi/format marks, and line/paragraph separators (`Cc`, `Cf`, `Zl`, `Zp`). Same check is on the token path so those names never become an `@` label. Added a regression test for U+2028, U+2029, NEL, U+202E, and ZWSP.

hardbeat920:
Thanks @elanchezhiyanr, this addresses my concern. Looks good to merge!

PR #70, opened 2026-09-04
Properly close tabs and files of archived or deleted sessions

kinsomicrote:
## What changed

- Archiving a session removes the tab
- Deleting a session removes open files

## Why

Before now, archiving a session left the tab in place, which is inconsistent with expectations. Deleting a session promoted open files to a session tab.

## UI

CleanShot 2026-09-04 at 20 41 58

## Checklist

- [x] I ran `npm run check`


kinsomicrote:
@hardbeat920 this PR fixes the bug I mentioned. Let me know if there are any changes I need to make.

hardbeat920:
Thanks for the fix @kinsomicrote

the workspace behavior makes sense and the lifecycle helper is well tested. 

Before merging, could we address two persistence races?

1. Archiving captures and persists open before stopping/flushing a busy turn, so recently buffered output may be lost. The save failure is also swallowed before the tab is removed.

2. Deletion runs before invalidating the turn and clearing pending writes, allowing a late upsertSession() to recreate the deleted database record.

Please stop/invalidate and flush first, persist the latest snapshot successfully when archiving, and serialize deletion after pending upserts. Integration tests covering archive/delete during an active turn would be valuable.

Thanks again 🙏

kinsomicrote:
@hardbeat920 I've updated the PR.

hardbeat920:
@kinsomicrote thank you! Merged :)

PR #72, opened 2026-09-04
feat: center code in diff

tcmarkfeld:
## Summary

- Center code in the diff viewer

Screenshot 2026-09-04 at 5 06 24 PM

## Verification

- `npm run build` passed
- tests passed

### DIFF
--- a/src/chrome/ProjectRail.tsx
+++ b/src/chrome/ProjectRail.tsx
@@ -12,7 +12,6 @@ import {
   File,
   Plus,
   Search,
-  Settings,
   Trash2,
 } from "./icons";
 import { useEffect, useMemo, useRef, useState, type MouseEvent } from "react";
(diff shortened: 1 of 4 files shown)


hardbeat920:
Thanks for the contribution @tcmarkfeld ! 

I like the diff-line centering change and am happy to keep that.
I’d prefer not to move “Check for updates” into the small icon beside Settings. 
The project sidebar isn’t cluttered enough yet to justify compressing it, and the icon-only treatment removes useful context such as the current version and update/download status.

Could you revert the sidebar update changes and leave just the diff-centering change in this PR?

Thanks again!

tcmarkfeld:
Updated @hardbeat920

hardbeat920:
@tcmarkfeld thank you. Merging :)

PR #76, opened 2026-09-04
Preserve composer draft when opening files from the explorer

kartava:
## What changed

`Composer` reports its draft to `SessionPane`, which keeps the latest text and passes it back as `initialDraft` when the composer remounts. Submit clears it so a sent message does not reappear.

## Why

Fixes a bug where a draft message disappears from the chat input when you open a file from the explorer. Type a message in a new session, click any file in the sidebar, and the input is empty.

The file opens in a split next to the session. That moves the composer from the empty view into the dock. React remounts it under the new parent and the typed text is lost. The same happens when a terminal opens in an empty session or the file pane closes.

Keeping the draft in `SessionPane`, which survives the split, fixes all three cases without touching the layout.

Not covered: attachments and the plan toggle still reset on that remount.

## UI

No visual change.

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/src/chrome/Composer.tsx
+++ b/src/chrome/Composer.tsx
@@ -169,6 +169,7 @@ type Props = {
   onSteerQueuedMessage?: (messageId: string) => void;
   onResumeQueue?: () => void;
   onOpenFile?: (path: string) => void;
+  onDraftChange?: (text: string) => void;
   children?: ReactNode;
 };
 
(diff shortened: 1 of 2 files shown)


hardbeat920:
thanks @kartava this was bugging me too and never got the chance to fix it myself 🙏

PR #80, opened 2026-09-05
Add Z.AI, MiMo, OpenRouter and NVIDIA providers

mondary:
## Summary

- Add first-class Z.AI, MiMo, OpenRouter and NVIDIA provider tabs.
- Route Z.AI and MiMo through their configured OpenCode providers.
- Add dynamic provider model catalogs, including OpenRouter public models.
- Add provider icons adapted from CodexBar.
- Filter known retired NVIDIA model entries.

## Validation

- Targeted catalog tests passing.
- Production build passing.
- Full test suite previously validated with 1173 tests passing.

## Notes

OpenRouter requires each user to configure their own OpenRouter credentials in OpenCode. Z.AI, MiMo and NVIDIA use credentials configured in OpenCode.


hardbeat920:
Thanks for putting this together @mondary  

I’m going to close this for now because we’re pausing all new provider integrations while we stabilize the providers already supported.

Although these providers reuse the OpenCode harness, exposing them as separate first-class providers still adds catalog, model-routing, credential, UI, and ongoing support complexity. 
The existing integrations currently have enough reliability and consistency issues that we don’t want to expand that surface area yet.

Once the provider layer is stable, we can revisit whether OpenCode-backed providers should be discovered generically or introduced individually through focused PRs against the latest main.

Thanks for the work i really appreciate it.

mondary:
Thanks for the thorough explanation — totally fair, and I'll keep my branch around for when the timing is right.

One clarification though: pointing users at the OpenCode provider isn't equivalent for everyone. OpenCode's auth routes traffic through its own gateway, but many of us already have API keys and active subscriptions elsewhere (Z.AI credits, OpenRouter, NVIDIA, etc.) and want to use our own billing, rate limits, and endpoints. That's exactly what native providers solve — so this isn't "more models", it's "bring your own endpoint".

Which leads me to a counter-proposal that aligns with your "discovered generically" idea rather than against it: a **user-defined custom provider entry** — name + base URL + model ID + API key, with a "test connection" button. It would:

- require zero catalog maintenance (model IDs are user-entered, never curated upstream),
- add no per-provider routing or credentials (one generic form, user-managed),
- cover every OpenAI-compatible provider without each becoming a first-class citizen.

That's strictly less surface area than the per-provider integrations you're pausing, and it unlocks the "not natively supported" case for everyone. Happy to reframe this work as a focused PR for just that, or to open an issue if you'd rather discuss the design first.

PR #83, opened 2026-09-05
Auto-collapse thinking and tool activity after turns

### LINKED ISSUE #79: Auto-collapse thinking + tool activity when a turn finishes (Codex-style compact transcript)
## Summary

After a turn completes, MonoCode often leaves the full **thinking / reasoning trail and tool-call activity** expanded above the final answer. That makes long runs hard to scan: the useful reply sits under a tall process log.

**Codex** (and several other agent UIs) keep the live trail visible *while* the turn runs, then **auto-collapse** it into a short summary once the final answer lands — so the transcript reads as:

1. compact work summary (e.g. duration / tool count), expandable on click  
2. the agent’s final response

I’d like MonoCode to support that same “settle → compact” behavior, **at least via a setting**.

## Current experience

During a long turn, MonoCode correctly shows thinking steps and tool work. When the agent finishes, that process block often remains fully (or mostly) expanded above the closing answer. Reviewing several turns means a lot of scrolling past process chrome to find what the agent actually said.

MonoCode already groups activity into **phases** that can fold (the old Zen transcript behavior, later made the default). That’s helpful, but the end state is still denser than Codex’s single collapsed summary — especially when there are many phases, reasoning paragraphs, or tool rows that stay easy to “see through” rather than tuck behind one line.

## How other products handle this

| Product | Pattern after the turn settles |
| --- | --- |
| **Codex** | Thinking + tool calls collapse into a short header (often duration-based, e.g. “Took Xm Ys”), with the final answer dominant; expand to re-read the trail. |
| **Cursor** | Work / thinking / exploration steps fold behind compact headers (`Worked for…`, `Explored N tools`, etc.). Conversation density controls how much detail stays inline, but the direction is clearly “compact by default after work.” |
| **Claude Code** | Tool rows and thinking lean toward collapsed / summary forms by default; users expand via transcript / toggles (`Ctrl+O`, thinking summaries, verbose / tool-output settings). Settings exist because people disagree on how much process to show. |

Shared idea: **live = open for monitoring; done = collapsed for reading**, with one click to reopen.

## Request

Please add **auto-collapse of thinking + tool activity when a turn finishes**, ideally with a setting so people who want the full trail left open can keep that.

### Suggested behavior

- **While running:** keep the current live view (phases open / pinned to latest activity is fine).
- **When the turn settles (or when the final prose answer starts):** collapse thinking + tool activity into a **single compact summary** above the answer — similar to Codex (duration and/or `N tool calls · M files edited`).
- **Click the summary** to expand the full trail again.
- **Setting** (name TBD), e.g. under Settings → General:
  - `Auto-collapse activity when turn completes` — on/off  
  - Optional later: `always expanded` / `collapse when done` / `collapse phases only` if you want finer control.

Even a single on/off that targets Codex-like compactness would cover the main ask.

### Nice-to-haves (non-blocking)

- Keep approval / failed / rejected tool rows more visible even when collapsed (errors shouldn’t hide behind the fold).
- If the user manually expands a finished group, don’t immediately re-collapse it on the next paint.
- Persist the preference across sessions.

## Why this matters

Coding-agent turns are getting longer. The final answer (and file diffs) is usually what you re-read; the process is for audit when something looks wrong. Auto-collapse keeps MonoCode’s transcript **scannable** without deleting any detail — just defaulting the finished trail to “one click away,” the way Codex already does.

jagadhis:
## What changed

Finished turns now fold all thinking and tool activity into one expandable summary above the final response. A persisted Settings → General toggle keeps the behavior optional, while live turns remain expanded and failed or rejected calls stay visible in the summary count.

## Why

Long agent runs currently leave several process phases above the final answer, making completed conversations difficult to scan. This implements the compact settled state requested in #79 without changing the live monitoring experience.

## UI

No screenshot attached. The change uses the existing transcript phase styling and adds the `Auto-collapse activity` toggle under Settings → General.

## Checklist

- [x] I ran `npm run check`

Closes #79

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Finished turns fold thinking and tool activity into one expandable summary above the final answer. The behavior is enabled by default and can be changed under Settings → General; failed or rejected calls remain counted in the folded header. In #79.
+
 ## [0.1.34] - 2026-09-05
 
 ### Added
--- a/src/lib/settings.test.ts
+++ b/src/lib/settings.test.ts
@@ -1,5 +1,6 @@
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import {
+  AUTO_COLLAPSE_ACTIVITY_DEFAULT,
   COMPOSER_RUNNER_DEFAULT,
   DIFF_VIEWER_DEFAULT,
   FOLLOW_UP_BEHAVIOR_DEFAULT,
@@ -7,6 +8,7 @@ import {
   KEYBINDINGS,
   LIVE_AGENTS_ENABLED_DEFAULT,
   loadComposerRunner,
+  loadAutoCollapseActivity,
   loadDiffViewer,
   loadFollowUpBehavior,
   loadGridArcadeEnabled,
@@ -14,6 +16,7 @@ import {
   loadNotesEnabled,
   NOTES_ENABLED_DEFAULT,
   saveComposerRunner,
+  saveAutoCollapseActivity,
   saveDiffViewer,
   saveFollowUpBehavior,
   saveGridArcadeEnabled,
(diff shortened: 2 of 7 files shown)


hardbeat920:
Thanks for this @jagadhis The auto-collapse looks good overall, but expanding the summary changes the transcript order. 

For example, tool A → progress message → tool B becomes progress message → tool A → tool B. 

We need to preserve the original activity/prose ordering when the summary is expanded?

jagadhis:
Fixed in 55fbd40. Expanded activity now renders each original activity group at its existing TurnItem position, so tool A → progress message → tool B keeps that order. The compact summary remains immediately above the final answer and still toggles the whole trail. I also reran npm run check successfully (1,249 web tests and 198 Rust tests).

hardbeat920:
I appriacete your efforts @jagadhis I tested this locally, and the collapsing causes too much layout and scroll shifting. Content jumps when tools are changing, when the turn settles, and when the summary is expanded. 

I think this needs to be handled in a different way. I would imagine, and that was the original idea but i haven't gotten the time to work on it, is to comabine as much of the tool calls withouth loosing context while streaming and once settleed to get the final answer. In its current form, the movement is too disruptive to merge

hardbeat920:
@jagadhis the latest release handles this. Give it a try and feel free to open another pr if you find issues with it. I will close this one for now. Thanks again

PR #86, opened 2026-09-06
Add app-wide UI scaling (Ctrl/Cmd + − 0)

xaccefy:
There was no scaling setting or config, so this adds browser-standard zoom controls.

## What
- **Shortcuts**: Ctrl/Cmd `+`/`=` zoom in, `-` zoom out, `0` reset. Works everywhere, including inputs and the terminal, like a browser.
- **Range**: 50–200% in 10% steps, persisted to localStorage (`monocode.uiScale`) and applied via CSS `zoom` on the document root (scales px-based Tailwind layouts, unlike root font-size).
- **Menus**: native macOS View menu (Zoom In / Out / Reset with accelerators) plus the custom MenuBar View menu on Linux/Windows.
- **Settings**: new Interface scale slider in Settings > Appearance, synced live via event, included in Restore Defaults.
- **Docs**: entries in the Settings > Keybindings list and the tabKeys comment.

## Files
- New: `src/lib/uiScale.ts` (+ `uiScale.test.ts` — clamping, step rounding, key mapping)
- Wired: `src/App.tsx` (global handler + menu-event listeners), `src/lib/appearance.ts` (apply on boot), `src/chrome/MenuBar.tsx`, `src/surfaces/SettingsView.tsx`, `src/lib/settings.ts`, `src-tauri/src/menu.rs`, `src/lib/tabKeys.ts`

## Checks
- `npm run check:web`: 122 files, 1256 tests pass (incl. 3 new uiScale tests); `tsc --noEmit` clean
- `cargo fmt --check` clean, `cargo clippy -- -D warnings` clean, `cargo test`: 199 pass


hardbeat920:
@xaccefy Thanks for adding this! 

A few things need fixing before we merge: 
- zooming throws off menu positions and pane resizing
- the scale slider gets stuck when using arrow keys
- zoom commands can run more than once with multiple windows open

Could you address those and resolve the merge conflicts?

xaccefy:
@hardbeat920 All four addressed in the latest push (rebased onto `main", conflicts resolved):

**1. Zoom throws off menus / pane resizing** — you were right, CSS `zoom` on the root breaks every `getBoundingClientRect`/`clientX` calculation in the app. I switched the engine to the native webview zoom (`Webview.setZoom`, i.e. Chromium page zoom / WebView2 zoom), which keeps all coordinate math consistent. Added the `core:webview:allow-set-webview-zoom` capability for it. Outside Tauri (plain `vite dev` in a browser) it falls back to CSS zoom so dev still works.

**2. Slider stuck on arrow keys** — the shared `Slider` had no `step` (default 1%), while the scale normalizes to 10% increments, so a ±1% arrow move rounded straight back. `Slider` now takes an optional `step` (default 1, existing sliders unchanged) and the Interface scale slider passes `step={10}`.

**3. Commands running more than once with multiple windows** — root cause was a double path on key presses: the native menu accelerators (`CmdOrCtrl+=` etc.) fired `zoom_in` via emit *on top of* the webview key handler applying the same step, so the focused window zoomed twice per press. I removed the accelerators from the native menu (items stay clickable and still work via emit); the webview key handler is now the single keyboard path on every platform. Each window applies a menu command exactly once to itself.

**4. Copilot: save-before-apply ordering** — fixed everywhere: every caller now persists via `saveUiScale` first and applies second, so `UI_SCALE_CHANGE_EVENT` listeners that read the store never see a stale value.

Checks on the rebased branch: `npm run check:web` 1267 pass, `tsc` clean, `cargo fmt` clean, `cargo clippy -D warnings` clean, `cargo test` 200 pass.

hardbeat920:
Thanks for the fixes @xaccefy  

One thing still needs adjusting, menu clicks are broadcast to every window, and each window increments the shared scale setting. Removing the keyboard shortcuts from the menu doesn’t fix that part. 
Could you make the menu action target just the focused window?

xaccefy:
Good catch — fixed. The menu dispatch used `app.emit`, so one click made every open window read-increment-write the shared setting (racing each other on top of it).

Zoom menu actions now target a single window: the focused one, falling back to the first visible window when nothing reports focus — which is the case to handle, since focus state is unreliable while the native menu is open. Broadcast is only the last-resort fallback when no windows exist at all. Keyboard zoom was already per-window (handled locally in the focused webview), and the custom MenuBar on Linux/Windows calls straight into its own window, so this was macOS-native-menu-only.

`cargo fmt` / `clippy -D warnings` / `cargo test` (200 pass) all green on the pushed commit.

PR #89, opened 2026-09-06
Add a shortcut to archive the focused conversation

kualta:
## What changed

Add Cmd+Shift+A (Ctrl+Shift+A on Windows/Linux) to archive the focused conversation through the existing archive lifecycle. List the shortcut in Settings → Keybindings.

A shared handler checks session focus and visible overlays before cancelling the event. Editors, diffs, terminals, dialogs, menus, and popovers retain their shortcut behavior, including when a popover leaves focus in the composer. Existing save, stop, and unsaved-file/terminal confirmations still apply.

## Why

Archiving currently requires the sidebar action or context menu. This makes it available directly from the keyboard while working in a conversation.

## Validation

- 8 focused key-mapping cases and 23 routing cases verify eligible archiving, untouched blocked events, open menus/popovers, and hidden overlays. Tests use the existing Node suite with lightweight DOM mocks and no new dependencies.
- Frontend suite on Node 20 with `vitest run --maxWorkers=2`: 1,313 tests passed; TypeScript passed.
- Rust formatting, Clippy, and 205 tests passed.
- Production frontend build passed. Native desktop interaction remains unverified because desktop automation access was unavailable.

## Checklist

- [x] Both halves of `npm run check` passed


hardbeat920:
This is great @kualta

One keyboard-routing issue needs fixing, the global handler calls preventDefault() and stopPropagation() before onArchiveFocusedSession checks whether an editor, diff, or terminal is focused. Those contexts don’t archive, but they still lose Cmd/Ctrl+Shift+A.
The overlay guard also misses ordinary popovers such as TabGroupMenu (role="menu"), so the underlying conversation can still be archived while that menu is open, despite the documented !overlay condition.

Could you centralize the archive eligibility check and run it before cancelling the event, covering session focus plus all open overlays/popovers?

kualta:
@hardbeat920 lmk if it looks good

hardbeat920:
Thanks @kualta the routing fix looks good now. 

One thing before merging, could you restore focused tests for the new shortcut? The final commit removes the archive-specific tabCommand and routing tests, so the current head doesn’t verify the key mapping or that blocked contexts leave the event untouched.

hardbeat920:
thanks @kualta this is amazing. merging

PR #90, opened 2026-09-06
Add prompt outline to session transcript

kartava:
## What changed

Adds a prompt outline to the session transcript: a stack of bars at the top right, one per user prompt, with the current one bright. Hover opens a list of prompts; a click scrolls the transcript to that prompt.

## Why

A long session gives no sense of where you are and no way back to an earlier prompt except scrolling. The outline does both.

Critical parts for review:

- `src/lib/promptOutline.ts` holds the active-row rule and its helpers as pure functions, covered by unit tests.
- `AgentTranscript` gains one attribute per user message and an `onRevealReady` prop, so a click can render a turn older than the rendered window before scrolling to it. The reveal uses `flushSync`.
- A row click sends a synthetic wheel-up to the transcript, so a streaming update does not pull the view back to the bottom. This relies on the transcript's existing wheel listener.
- On panes narrower than 62rem the transcript column keeps a right gutter for the widget.

## UI
Before:
Screenshot 2026-09-06 at 19 29 12
After:
Screenshot 2026-09-06 at 19 29 57
Screenshot 2026-09-06 at 19 30 04
Screenshot 2026-09-06 at 19 30 14

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/src/chrome/PromptOutline.tsx
+++ b/src/chrome/PromptOutline.tsx
@@ -1,0 +1,352 @@
+import {
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+  type MouseEvent as ReactMouseEvent,
+  type RefObject,
+} from "react";
+import {
+  activePromptId,
+  barWindow,
+  promptBlocks,
+  promptLabel,
+  type OutlineAnchor,
+  type OutlineBand,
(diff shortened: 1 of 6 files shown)


hardbeat920:
@kartava thank you. Nice feature addition

PR #91, opened 2026-09-06
fix(harness): show live CLI models in the picker

Warexpor:
## Why
Installed harness CLIs already list models the static picker catalog does not include. Opening that harness tab should force a live probe and show those models under the harness the user has installed.

## Scope
- `refreshHarnessCatalogs` accepts `{ force }` so a live overlay does not skip a re-probe.
- `ModelPicker` force-refreshes the visible harness tab.
- Cursor and Grok catalogs try CLI before ACP.
- Grok discovery returns `[]` when CLI and ACP miss, so synthetic `fallbackGrokModels` no longer marks a live catalog.
- Parser and registry tests cover muse-spark lines, Cursor family grouping, and force refresh.

Out of scope: Windows `check-rust.mjs`, durable/persisted catalogs, boot-probing every installed CLI.

## Tradeoffs
CLI list fully replaces static rows for that harness. Cursor collapses effort/fast variants into one family row with settings. A brief static flash can appear before the overlay lands.

## Blast Radius
ModelPicker rows for Cursor and Grok change after tab open. Session boot refresh stays scoped to harnesses already in the window. Other harness catalogs gain force refresh for free when their tab opens.

## Verification
- `npx vitest run src/lib/harness/cursorCatalog.test.ts src/lib/harness/grokProtocol.test.ts src/lib/harness/registry.test.ts` passed.
- `npm run check:web` passed (1298 tests).
- Manual: `cursor-agent --list-models` and `grok models` on this machine match parser fixtures for sol/luna/auto and muse-spark.


hardbeat920:
Thanks for putting this together @Warexpor 

After reviewing it, I don’t think we want to take this direction right now. Monocode already loads live model catalogs, and we intentionally avoid repeatedly probing a harness once a live catalog has landed because of the additional process and resource overhead.

The PR also includes the fork’s broader history (93 commits and changes across 57 files) which makes the intended change difficult to review and merge independently.

I’m going to close this for now, but I really appreciate the time and thought you put into it.

PR #93, opened 2026-09-06
Let wrapped inline code grow and keep list markers beside it

kartava:
## What changed

Inline code pills now grow with their text instead of sticking to a fixed height. Lists put the number or bullet in a gutter, so a pill that wraps stays next to its marker.

## Why

A long inline code span inside a numbered list looked broken. The number sat alone on its own line, and the background was a thin band cutting through two lines of text. The pill could not wrap because it is an `inline-flex` box, and `h-6` locked its height.

## UI

Before: 
Screenshot 2026-09-06 at 20 10 09
Screenshot 2026-09-06 at 20 17 28

After:
Screenshot 2026-09-06 at 20 10 18

Screenshot 2026-09-06 at 20 18 08

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -185,6 +185,13 @@ input,
     color-mix(in srgb, var(--color-content) 20%, transparent);
 }
 
+.agent-markdown [data-streamdown="ordered-list"],
+.agent-markdown [data-streamdown="unordered-list"] {
+  list-style-position: outside;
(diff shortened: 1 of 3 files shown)


hardbeat920:
@kartava thanks this is great :)

PR #94, opened 2026-09-06
Fix terminal focus loss after changing directory

### LINKED ISSUE #88: Bug after run $ cd ..
### MonoCode version

0.1.34

### Operating system

macOS

### OS version

macOS 26.3

### Provider CLI

Claude Code

### What happened

After open the terminal below with "⌘J" and running something like "$ cd .." or "$ cd  app/" it won't let you type, and forces you to press the terminal again to be able to type.

### Steps to reproduce

1. ⌘J
2. cd  ~
3. try to type something

bytehello:
## Summary

- keep the xterm/PTY lifecycle tied to terminal identity rather than live cwd metadata
- prevent OSC 7 cwd reports from tearing down and recreating an active terminal
- preserve keyboard focus after `cd` updates the terminal tab metadata

Closes #88

## Testing

- `npm run check:web` (124 test files, 1282 tests, and `tsc --noEmit` passed)
- `git diff --check`
- `npm run check:rust` was not run because the local environment does not have `cargo` installed; this change is limited to the React effect dependency list.

### DIFF
--- a/src/surfaces/TerminalView.tsx
+++ b/src/surfaces/TerminalView.tsx
@@ -320,7 +320,7 @@ export function TerminalView({ id, cwd, active, onMetaChange }: Props) {
       termRef.current = null;
       spawned.current = false;
     };
-  }, [id, cwd]);
+  }, [id]);
 
   // Identity-stable: the callers pass an inline arrow, so depending on the
   // prop itself would tear down and re-arm the poll — and re-fork `ps` — on


hardbeat920:
this is great @bytehello! couldnt reproduce maybe because my shell doesnt emit OSC 7 directory that triggers the bug
Merging once ci passes

PR #101, opened 2026-09-06
Add Antigravity, newer model catalogs, sub-agents, and app mark

HaydernCenterpoint:
## Summary
Splits the recent MonoCode work into five focused commits:

1. **Antigravity provider** — detect `agy`, load its live catalog, and drive turns through the harness registry.
2. **Newest-first catalogs + searchable pickers** — rank models by version, default to the latest family match, replace native `<select>` with a searchable popover, and add OpenCode's Free filter.
3. **Sub agents settings** — pin helper-agent models per provider, with Claude Agent-tool wiring.
4. **Cursor on Windows** — find `cursor-agent` under LocalAppData and accept `.cmd` shims.
5. **App mark** — show the terminal-grid logo in the title bar and regenerate store/app icons.

## Test plan
- [x] Settings → Providers shows Antigravity when `agy` is installed, with Gemini/Claude/GPT-OSS models.
- [x] Provider model dropdowns are searchable; OpenCode Free filter reduces the list to zero-cost models.
- [x] Settings → Sub agents can inherit or pin a helper model.
- [x] Cursor CLI is detected on Windows when installed via LocalAppData.
- [x] Title bar shows the MonoCode mark next to the Development badge.


hardbeat920:
Thanks for putting this together @HaydernCenterpoint 

I’m going to close this PR because it combines several independent changes, which makes it difficult to review and land safely.

We’re also pausing new providers and further provider expansion until the existing integrations are stable. 
Some of the other changes here may still be useful, but they should be proposed separately as small, focused PRs against the latest main. 

Please open an issue or check with us before reworking them, since the project is moving quickly and some areas may already be changing.

Thanks for the effort!

PR #103, opened 2026-09-07
Filter the Linear inbox by team and project

ruben-carrera:
## What changed

The inbox filter menu now has **Teams** and **Projects** sections on the Linear tab. Teams reuses the existing hidden-team list from Settings; Projects is new, and needs one extra field on the issues query.

## Why

Linear issues could not be narrowed by where they live. Teams were only reachable from Settings → Linear Teams, which is a long way from the list you are looking at, and projects were not fetched at all — so on a workspace with a few active projects the inbox was all-or-nothing.

Two deliberate asymmetries, since the sections look identical but are not:

- **Teams stays fetch-level.** It reads and writes the same `monocode.linearHiddenTeams` the Settings pane uses, so the two stay in sync and unchecking a team still narrows the GraphQL `IssueFilter` rather than just the rendered list. The roster comes from `listLinearTeams()` and not from the fetched issues on purpose: hiding a team drops its issues, so a derived list could never offer that team back.
- **Projects is client-side.** Pushing it into `IssueFilter` would shrink the result window and let unrelated issues slip in under the fetch cap, so the filter hides already-fetched issues instead. A trailing "No project" row (a sentinel id, never a real Linear id) makes unassigned issues as hideable as any project — without it, narrowing to one project still leaves every loose issue on screen.

Hidden projects are deliberately not pruned. `pruneInboxFilters` drops stale local projects against the rail's known paths, which is safe because that list is stable; the Linear project list is derived from whatever the current fetch returned, so pruning against it would silently clear the filter whenever a hidden project happened to have no issues in the window.

Two smaller things that fell out: `Clear filters` also clears hidden teams (otherwise the funnel icon stays lit with no way to reset it from the menu), and `LinearSettings` now follows `LINEAR_CHANGE_EVENT` — it seeded its checkboxes from `localStorage` once, so toggling a team in the inbox left an open Settings pane showing stale state.

Known limitation: the project list only contains projects present in the fetched issues, so a project whose issues all fall outside the window won't offer a row. Fixing that properly means a separate projects query, like `listLinearTeams` does for teams — happy to add it if you'd rather have it upfront.

No CHANGELOG entry, since `Unreleased` entries look like they're written at release time with the PR number attached — glad to add one if you'd prefer it in the PR.

## UI

The Linear filter menu goes from Assigned to me · Status · Time to Assigned to me · Status · Time · Teams · Projects. Happy to attach a screenshot — flagging that I have not exercised this against a live workspace beyond my own, so a second look at the team list matching Settings is worthwhile.

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/src-tauri/src/linear.rs
+++ b/src-tauri/src/linear.rs
@@ -58,6 +58,8 @@ pub struct LinearIssue {
     pub repo: String,
     pub team_id: String,
     pub team_name: String,
+    pub project_id: String,
+    pub project_name: String,
     pub project_path: String,
 }
 
@@ -247,6 +249,7 @@ query InboxIssues($first: Int!, $filter: IssueFilter) {
       updatedAt
       state { name type }
       team { id key name }
+      project { id name }
       labels { nodes { name color } }
       assignee { name displayName avatarUrl }
     }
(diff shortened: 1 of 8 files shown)


hardbeat920:
Thanks @ruben-carrera this look great. Merging

PR #104, opened 2026-09-07
Key project appearance by path instead of folder name

varienos:
## What changed

Project appearance — rename/label, color, custom color, logo, mascot — and **Remove project data** were keyed by the project's **folder name**. They are keyed by the project's full path now, with a one-time migration for settings people already have.

## Why

I ran into this while adding two checkouts that happen to share a folder name:

```
~/cortex-finance/agentbase
~/cortex/agentbase
```

Both resolved to the key `agentbase`, so the two rail entries were really one record:

- Renaming one renamed the other.
- They shared color, custom color, logo and mascot.
- Removing one project's data wiped the other project's settings with it.

The cause is that `ProjectRail`, `Sidebar`, `Composer`, `ComposerRunner`, `InboxView`, `NotesView`, `SettingsView` and `projectData.ts` all passed `projectName(path)` — the last path segment — as the key into the `monocode:tab-group:*` records.

### How it is fixed

- **`projectKey(cwd)`** in `src/lib/paths.ts` returns `pathKey(cwd)`: the normalized full path, case-folded on Windows. Every read and write of the appearance stores goes through it.
- **`projectName()` is untouched in meaning.** It stays the display name, and it stays the hash seed for the *default* color and mascot, so existing installs keep the exact colors and mascots they have today. Only the override lookup moved.
- **Migration** (`tabGroups.ts`, storage version `2`) copies each folder-name entry onto every remembered project that carries that name. Both rails look identical right after the update and only diverge once one of them is edited. A name that no known project claims — a project pushed out of the 20-slot recents cap, say — leaves the pass unfinished, and later launches retry, so reopening that project still restores its label rather than losing it for good.
- **Logo filenames** (`src-tauri/src/project_logo.rs`) are derived from the key, and sanitizing alone cannot keep full paths apart: `a-b/c` and `a/b/c` both collapse to `a-b-c`, which is exactly the shape of the pair that started this. Every stem now carries an FNV-1a hash of the key and keeps only as much readable tail as fits the filesystem limit. That also stops `proj` from matching `proj.old.png` when a logo is replaced.
- **`forget_logo_file`** removes the image a project no longer points at. A logo saved before this change lives under a stem nothing derives anymore, so the stored path is the only handle left to it; anything outside the logo directory is ignored.

## Tested

Reproduced with the exact pair above, then confirmed in the running app on macOS after the change:

- Renaming `cortex-finance/agentbase` to "Finance" leaves `cortex/agentbase` as "agentbase". Color, logo and mascot are independent, and removing one project's data no longer touches the other.
- Labels and colors saved before the update survive it, and a project reopened after dropping out of recents gets its label back.

New tests cover the regression rather than the happy path:

- `src/lib/tabGroups.test.ts` — the same-name collision, the migration, the retry for a project that is not remembered yet, and a Windows path that differs only in case.
- `src-tauri/src/project_logo.rs` — the two path shapes that used to sanitize to one logo stem, the sibling-prefix case, and the length cap.

`npm run check` is green: 1288 vitest tests, `tsc --noEmit`, `cargo fmt`, `cargo clippy -D warnings`, 207 cargo tests.

## UI

No visual change by design. Defaults still hash from the folder name, so nothing shifts for existing users; the only behavior change is that two same-named projects can finally be renamed and styled independently.

## Out of scope

Tab-group *joining* (`workspaceTabGroups.ts`) still scopes by folder name, so tabs from two same-named checkouts can merge into one group. Same root cause, but it touches no persisted state and it is a different surface, so I left it out to keep this to one thing. Happy to send it separately if you want it.

## Checklist

- [x] I ran `npm run check`

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -319,6 +319,7 @@ pub fn run() {
             window_transfer::take_window_transfer,
             project_logo::save_project_logo,
             project_logo::remove_project_logo,
+            project_logo::forget_logo_file,
         ])
         .build(tauri::generate_context!())
         .expect("error while building MonoCode");
--- a/src-tauri/src/project_logo.rs
+++ b/src-tauri/src/project_logo.rs
@@ -17,12 +17,29 @@ fn project_logos_dir(app: &AppHandle) -> Result<PathBuf, String> {
     Ok(dir)
 }
 
+/// Projects are keyed by their full path, which sanitizing alone cannot keep
+/// apart: `a-b/c` and `a/b/c` both collapse to `a-b-c`, and a deep checkout can
+/// outgrow the filesystem's name limit. So every stem carries a hash of the
+/// whole key and keeps only as much of its readable tail as fits.
+const MAX_STEM_CHARS: usize = 96;
+/// `-` plus the hex hash.
+const HASH_CHARS: usize = 17;
+
+fn fnv1a(value: &str) -> u64 {
+    let mut hash: u64 = 0xcbf2_9ce4_8422_2325;
+    for byte in value.as_bytes() {
+        hash ^= u64::from(*byte);
+        hash = hash.wrapping_mul(0x1000_0000_01b3);
+    }
+    hash
+}
+
 fn sanitize_project_key(project: &str) -> String {
     let trimmed = project.trim();
     if trimmed.is_empty() {
         return "project".into();
     }
-    trimmed
+    let safe: String = trimmed
         .chars()
         .map(|ch| {
             if ch.is_ascii_alphanumeric() || ch == '-' || ch == '_' || ch == '.' {
@@ -31,7 +48,14 @@ fn sanitize_project_key(project: &str) -> String {
                 '-'
             }
         })
-        .collect()
+        .collect();
+    let head = MAX_STEM_CHARS - HASH_CHARS;
+    let readable: String = if safe.chars().count() <= head {
+        safe
+    } else {
+        safe.chars().skip(safe.chars().count() - head).collect()
+    };
+    format!("{readable}-{:016x}", fnv1a(trimmed))
 }
 
 fn logo_stem(project: &str) -> String {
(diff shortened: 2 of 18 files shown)


hardbeat920:
Thanks @varienos this fixes a real same-folder-name collision!

Before merging, please address:
1. Migrated same-name projects share the old logo file. Changing or clearing one deletes the file used by the other. Only delete it when no other project references it, and add a regression test.
2. The FNV-1a multiplier should be 0x100_0000_01b3, not 0x1000_0000_01b3.

thanks again 🙏

varienos:
Thanks for the review — both were real, both fixed in `b829992`.

**Shared logo file:** you're right. The migration points both same-named projects at the one file saved under the old folder-name key, and `clearProjectLogo` / `pickAndSetProjectLogo` deleted it unconditionally, so clearing or replacing either one pulled the image out from under the other. There is now a `droppableLogoFile(logos, project, keep?)` helper that returns the stored path only when no other project still references it, and both call sites go through it — the file survives until the last project lets go.

**FNV-1a multiplier:** good catch. I had written twelve hex digits instead of eleven, so the constant was `17592186044851` rather than the FNV-1a prime `1099511628211`. Fixed. The hashed stems only exist in this PR, so no installed copy has files written under the old value and nothing needs migrating.

**Regression test:** new `src/lib/projectLogos.test.ts`, six cases — the shared file is kept and `forget_logo_file` is never invoked, the file is removed once the last project drops it, a replace that lands on the same path keeps the file, and a project with no logo has nothing to drop. I confirmed it actually guards the behaviour rather than just passing: restoring the unconditional delete turns two of them red, and they go green again once reverted.

`npm run check` is green — 1294 vitest tests, 207 cargo tests, `cargo fmt` and `cargo clippy -D warnings` clean.

hardbeat920:
@varienos some windows tests failed if you can check

varienos:
Checked it — the Windows leg is a flaky test rather than a Windows problem, and it does not come from this PR.

In that run only `check (windows-latest)` actually failed; macOS and Ubuntu are marked `cancelled`, killed by the matrix fail-fast once Windows went red. The failure is `src/surfaces/pacmanArcade.test.ts > detours to swallow the provider logos`, asserting `expected 3 to be greater than or equal to 4`.

**What is going on.** That test plays two minutes of the arcade and counts how many provider logos pac-man swallows, but the maze layout, the logo drops and the mascot releases all come off `Math.random`. I sampled the distribution locally: across 1500 unseeded runs the count averages 7.8 against a required 4, yet 11 of them land under it — about a 0.7% failure rate on every CI run, which is exactly the 3 Windows reported. The tail is self-reinforcing rather than a bug: a new logo only drops once the previous one is gone, so on an unlucky board where pac-man is walled off, or fenced in by hunting mascots, the logo burns its full 12s lifetime, which both scores a miss and pushes the next drop back. Two of those in a row and the count collapses.

Nothing about it is platform-specific, and this PR touches no arcade or session code, so the same failure is reachable on `main` — it just happened to land on the Windows runner this time.

**Fix** (`31e7588`): the test file now pins `Math.random` to a seeded mulberry32 stream in `beforeEach` and restores it in `afterEach`, so the same board plays out on every runner. To be sure the seed was not just a lucky draw hiding a weak assertion, I ran the full 21-test file against 40 different seeds — all green — and confirmed that the seeds my sampling had flagged reproduce the exact failure from the Windows log. `ARCADE_SEED` overrides the seed if you ever want to sweep other boards.

`npm run check` is green: 1294 vitest tests, `tsc --noEmit`, `cargo fmt`, `cargo clippy -D warnings`, 207 cargo tests. Safe to re-run CI whenever you like.

One note in the interest of the "no unrelated changes" box I ticked: this genuinely is unrelated to the subject of the PR. I folded it in because it was blocking the checks here, but I am happy to lift it into its own PR and rebase this one on top if you would rather keep the two apart.

hardbeat920:
Sorry for the confusion @varienos the CI failure was unrelated. 
Thanks for investigating and fixing the flaky test too. 
The changes look good!

PR #129, opened 2026-09-08
Add a setting to archive conversations with the close shortcut

kualta:
## What changed

Add an opt-in **Archive on close shortcut** toggle in Settings → General. When enabled, Cmd+W / Ctrl+W uses the same archive action as Cmd+Shift+A / Ctrl+Shift+A. The default remains normal pane closing.

Route the keyboard shortcut and macOS native close command through the existing centralized archive eligibility checks. Editors, diffs, terminals, other app surfaces, and visible overlays/popovers remain protected; blocked keyboard events stay untouched. Native commands also require the receiving window to have focus.

## Why

Let users archive finished conversations with their usual close shortcut while keeping them available in Archive.

## UI

The toggle appears in General settings; the keybindings reference documents both modes. Verified the real settings component in a browser preview: default off, enabling persists across reload, and disabling restores normal closing. Native desktop interaction remains unverified because desktop automation access was denied.

## Validation

71 focused settings, key mapping, and archive routing tests pass, including native command focus checks. The full frontend suite passes (1,404 tests), as do TypeScript and the production build. Rust formatting, Clippy, and 218 Rust tests pass on the identical native-code baseline. No dependencies added.

## Checklist

- [x] I ran `npm run check` (web and Rust halves separately; Vitest limited to two workers)

## Summary by CodeRabbit

- **New Features**
  - Added an “Archive on close shortcut” setting in General settings.
  - When enabled, the close-pane shortcut archives the focused session instead of closing it.
  - The shortcut remains available for closing panes when the setting is disabled.
  - Archive-on-close actions now respect window and workspace focus.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -328,6 +328,7 @@ import { linearIssueDetails, peekLinearIssueDetails } from "./lib/linear";
 import {
   loadLiveAgentsEnabled,
   loadNotesEnabled,
+  loadArchiveOnClose,
   loadDiffViewer,
   loadFollowUpBehavior,
   loadSettingsSection,
@@ -2897,7 +2898,7 @@ export default function App({
   );
 
   const onArchiveFocusedSession = useCallback(
-    (event: KeyboardEvent) => {
+    (event?: KeyboardEvent) => {
       archiveFocusedSession(
         event,
         {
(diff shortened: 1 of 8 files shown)


hardbeat920:
Thanks for putting this together @kualta  

I’m planning to support general user-configurable keybindings, so I don’t want to add a dedicated archive-on-close setting that would become a special case and likely need replacing later. 

I’m going to close this in favor of approaching archive shortcuts through the broader keybinding system. 

Thanks again for the contribution.

PR #132, opened 2026-09-08
Hide the sidebar update control unless an update is available

felipeorlando:
## Summary
- The sidebar update probe now lives in `SidebarUpdateFooter`, which only renders when `isSidebarUpdateActionable` is `available` or `downloading`. Idle, checking, and failed probes stay silent, so the wrapper padding disappears too.
- `SidebarUpdate` is presentational: the dead “Check for updates” / “Checking…” labels and the `RefreshCw` action are gone. Settings and app-menu manual checks keep their own snapshots.
- Footer/Settings spacing swapped (`footer` is `p-2 pb-0`, Settings is `p-2`) so Settings does not sit flush against the session list when the update control is hidden.

## Summary by CodeRabbit

- **New Features**
  - Sidebar updates display available and downloading states, including installation progress.
  - Post-install update information remains available without showing an unnecessary update row.

- **Bug Fixes**
  - Prevented multiple installation attempts when the install action is clicked repeatedly.
  - Standardized spacing around the Settings action in the sidebar.

- **Changes**
  - Removed the manual “Check for updates” action from the sidebar.

### DIFF
--- a/src/chrome/ProjectRail.tsx
+++ b/src/chrome/ProjectRail.tsx
@@ -488,7 +488,7 @@ export function ProjectRail({
             onOpenWhatsNew={onOpenWhatsNew}
             onDismissUpdate={onDismissUpdate}
           />
-          <div className="flex shrink-0 flex-col gap-px p-2 pt-0">
+          <div className="flex shrink-0 flex-col gap-px p-2">
             <RailAction
               label="Settings"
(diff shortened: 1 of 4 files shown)


hardbeat920:
thanks @felipeorlando looks great! People started complaining about this button so you are doing everyone a favor 😁

PR #133, opened 2026-09-08
Make toggle sidebar hide the full sidebar

RayenTellissy:
Cmd/Ctrl+B and View > Toggle Sidebar previously only collapsed the project rail (duplicating the rail's own toggle). This wires the command to hide the entire sidebar instead.

- New persisted `monocode.sidebarOpen` flag in `appearance.ts` with load/save helpers
- `App.tsx` tracks `sidebarOpen` state and passes it to `Sidebar` via the previously hardcoded `open` prop
- Project rail visibility now also respects `open`, with settings still forcing the rail slot visible since they render there

Rail collapse via its own button is unchanged. Menu item, accelerator, and keybindings entry already existed, so no changes there.

Verified with `vitest` (1397 passing) and `tsc --noEmit`.

## Summary by CodeRabbit

* **New Features**
  * Sidebar open/closed state is now saved and restored between sessions.
  * The project rail and settings rail are hidden when the sidebar is closed.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,8 +21,10 @@ import { UsageFooter } from "./chrome/UsageFooter";
 import { useProjectBranches } from "./hooks/useProjectBranches";
 import {
   loadProjectRailOpen,
+  loadSidebarOpen,
   loadSidebarTabOrder,
   saveProjectRailOpen,
+  saveSidebarOpen,
   type SidebarTabId,
 } from "./lib/appearance";
 import { HAS_NATIVE_GLASS, IS_MAC } from "./lib/platform";
@@ -601,6 +603,7 @@ export default function App({
     [],
   );
   const [projectRailOpen, setProjectRailOpen] = useState(loadProjectRailOpen);
+  const [sidebarOpen, setSidebarOpen] = useState(loadSidebarOpen);
   const tabCloseScope = "project" as const;
   const currentProjectDock = findProjectTerminal(projectTerminals, projectCwd);
   const dockVisible = !!currentProjectDock?.open;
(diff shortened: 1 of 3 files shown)


hardbeat920:
i am not sure that we should collapse the sidebar that holds the sessions as well... maybe it can be a different keybinding but not Cmd/Ctrl+B... what do you think ?

RayenTellissy:
yeah you can do a keybinding, but i think there has to be the option to just have the chat. some people like that experience (me included). every gui i use has that option

nulljosh (commented):
Logic checks out — this fixes a real bug where the sidebar toggle was mutating projectRailOpen instead of its own state, and correctly gates the git-status polling/keyboard shortcuts (useGitFileStatuses, useProjectDiffStats, sourceControlActive) on the new real `open` value instead of a hardcoded `open` shorthand that was always true.

One nit: the new lines in the onToggleSidebar callback (App.tsx) drop semicolons, inconsistent with the rest of the file which uses them (prettier ^3.9.6 is a project dependency) — worth a `prettier --write` pass before merge.

hardbeat920:
> yeah you can do a keybinding, but i think there has to be the option to just have the chat. some people like that experience (me included). every gui i use has that option

i understand and its a valid point the difference in other guis is that they render one sidebar where monocode has 2 in reality...

i wouldn't make this the default behaviour to collapse both... what would be an alternative option here @RayenTellissy

RayenTellissy:
it doesn't have to be default behavior, each sidebar can have it's own keybind to hide/show it

hardbeat920:
yeah that works with me. keybinding would be perfect just not more panel close buttons cause it would get confusing

PR #139, opened 2026-09-09
Fix blocking Codex approvals and missing user prompts

yankawai:
## What changed

Fix Codex access and server-request handling so Full Access can complete explicit escalation requests, questions and confirmations reach the existing MonoCode UI, and subsequent input requests generate notifications.

## Why

**Severity: high — this blocks a core application workflow.** A user can select Full Access and still have an authorized operation rejected before a usable confirmation appears. Separately, questions were answered with empty responses without ever appearing in the chat. Together these failures make ordinary inspection and development tasks stop or continue with missing information, while leaving the user unable to respond.

Please consider this for the next patch release: the impact is functional, not cosmetic, and repeatedly interrupts normal use of the application.

Investigated against MonoCode `0.1.40`, upstream commit `8dfea58`, and the Codex `0.153.4` app-server schema. A recorded Full Access turn used `approvalPolicy: "never"` and `danger-full-access` and returned:

```text
approval required by policy, but AskForApproval is set to Never
```

The failure above is a client policy mismatch; it should not be described as the user declining a dialog or an AI reviewer rejecting the operation.

| Failure | Cause and fix |
| --- | --- |
| Explicit escalation fails in Full Access | `never` rejects the request before MonoCode's existing automatic approval handler can answer. Use `on-request` with the same `danger-full-access` sandbox and `user` reviewer. Command, file and permission requests retain the Full Access automatic approval behavior. |
| Questions never appear | `item/tool/requestUserInput` returned `{ answers: {} }` immediately and the Codex adapter lacked `respondQuestion`. Route questions to the shared form and return selected labels/free text under the provider's question IDs. Honor `isBlocking`: optional questions auto-skip after 60 seconds of grace plus a 60-second countdown, matching Codex 0.153.4. Interaction keeps the question open; queued requests receive their own deadline when shown. Blocking and legacy requests continue waiting for a user decision. |
| Concurrent or resolved requests leave missing/stale UI | Queue questions; register pending approvals before publishing them; clear pending requests on server resolution, turn completion, cancellation, process exit and session stop. Ignore late answers and avoid duplicate responses. |
| MCP confirmation is unusable | Handle empty forms and single-Boolean MCP confirmations through the existing Allow/Deny controls. Allow returns `accept` with the schema's Boolean property set to `true`; Deny returns `decline` with null content. Show the field label/description, and keep consent explicit even in Full Access. Unsupported forms and unknown methods now produce visible notices and valid cancellation/error responses. |
| Later requests in one session do not notify | Notification deduplication used only the session ID. Carry the triggering approval/question ID into notification rendering, so a new approval cannot repeat an older question. Coalesce each render by session while marking only the selected request as seen; remaining requests stay eligible after subsequent updates or resolution. |
| External-clock turns fail with an unsupported-method error | Support Codex `currentTime/read` with `currentTimeAt` in whole Unix seconds, including requests during thread start/resume before the live session is bound. Keep the experimental API required by `collaborationMode` and Plan. |
| Access selection and active policy disagree | Bind queued mode changes when their turn actually starts. Explain in the existing picker that changes apply to the next turn; stopping and resending applies the selection immediately. |
| Plan silently selects the AI reviewer | Preserve the selected access mode's reviewer instead of forcing `auto_review` for every Plan turn. Keep Plan's non-escalating read-only sandbox. |

### Scope and compatibility

- Supervised still waits for an explicit decision; Auto still uses its AI reviewer. Full Access does not fabricate answers to user questions or MCP consent requests.
- Notification opt-in, OS permission and foreground suppression are preserved.
- Unsupported complex MCP forms/browser sign-in and secret input remain unsupported. They now report the limitation visibly; the shared transcript form is not used to collect secrets. Nonfatal notices do not clear another pending question or end an active turn.
- No new dependencies, database changes, global settings changes, or new approval UI components.

## UI

Reuses the current `QuestionForm`, transcript approval controls, `ApprovalToasts`, `Popover` and theme styles. Adds the access-change timing hint and an optional-question countdown within the existing form. The Auto description accurately says the reviewer can approve or deny.

The initial UI changes were verified in a local Safari render with test data: Allow, option selection, Continue, question/toast removal, and the access picker. DOM integration tests also click the real components and inspect the resulting app-server responses. Native notification dispatch is covered with a mocked native bridge; OS banner delivery and a newly packaged desktop release were not tested.

## Validation

- The initial regression tests failed in 15 cases on the unmodified upstream implementation. The two maintainer-reported cases reproduced as 7 further failures before this update: 4 external-clock cases and 3 notification content/delivery cases.
- `npm run check`: **1,487 web tests, 218 Rust tests**, TypeScript, Rust formatting and Clippy all pass.
- `npm run build`: passes.
- The remaining MCP Boolean and non-blocking question cases fail on the previous PR head and pass with this update. Tests cover both MCP decisions, optional expiry, preserved user answers, queued deadlines, cancellation/cleanup, and the actual form/toast/notification flow.
- Optional-question timing follows the [Codex 0.153.4 TUI policy](https://github.com/openai/codex/blob/rust-v0.153.4/codex-rs/tui/src/bottom_pane/request_user_input/mod.rs).
- Coverage includes fresh/resumed Full Access turns, supervised denial, Plan questions, queued mode changes, concurrent questions, cancellation/resolution, malformed and secret input, MCP confirmation/cancellation, clock reads during fresh/resumed setup and active Plan/default turns, concurrent notification content and deferred delivery, and real UI button-to-response routing. Notification integration tests now mount the same hook used by App and verify dispatch from actual harness events.

## CI reliability

A separate CI commit limits Linux dependency installation to the runner's official Ubuntu source file, avoiding failures in unrelated preinstalled repositories (the reported failure was Google Chrome `Hash Sum mismatch`). Package signature/hash checks remain enabled. macOS, Linux, and Windows jobs now finish independently. The setup helper retains its existing default for local development.

## Checklist

- [x] I ran `npm run check`
- [x] This PR is focused on the Codex access and input-request path

## Summary by CodeRabbit

- **New Features**
  - Added interactive Codex questions with custom responses, option selection, and MCP confirmations.
  - Optional questions now show auto-resolution countdowns and remain open when users begin responding.
  - Improved notifications for pending approvals and questions across sessions.
  - Access changes during a turn indicate they apply to the next turn, with guidance for applying them immediately.
  - Full-access mode can request explicit approval for actions.

- **Bug Fixes**
  - Improved cleanup and cancellation of pending questions and approvals.
  - Server-request errors are now surfaced correctly.

- **Documentation**
  - Clarified automatic review mode guidance.


hardbeat920:
@yankawai thanks for the thorough fix and test coverage 

I pulled the branch; the targeted tests and TypeScript check pass. 

I found two issues worth addressing before merge:
1. MonoCode initializes Codex with experimentalApi: true, but currentTime/read falls into the new -32601 Unsupported method branch. This is a valid Codex 0.153.4 experimental request, and rejecting it can terminate turns that use an external clock. Please avoid unsupported experimental capabilities.
2. The new per-request notification tracking does not pass the triggering request into notifySession. If an approval arrives while a question is pending, notificationText prefers the existing question, so the banner repeats the old question instead of describing the new approval. Because all concurrent keys are marked as seen, the remaining approval may never receive its own notification after the question is resolved.

The main approval/question lifecycle changes otherwise look well structured. 
I’d be happy with this after those two cases are covered.

hardbeat920:
Thanks for the updates @yankawai the clock, notification, and CI fixes look good.

Two issues from my earlier review still appear unresolved:
1. Standard MCP confirmations with a required Boolean field are still treated as unsupported and automatically cancelled.
2. requestUserInput still ignores isBlocking, so non-blocking questions can wait indefinitely.

Otherwise, the implementation looks solid and CI is green.

yankawai:
@hardbeat920 both remaining cases are addressed in `0e49c3b`: required-Boolean MCP confirmations now use the existing Allow/Deny controls, and `requestUserInput` honors `isBlocking`, with optional-question expiry and interaction handling covered by regression tests.

CI passes on macOS, Ubuntu and Windows. `npm run check` passes with 1,487 web tests and 218 Rust tests; the build also passes.

Could you please take another look and prioritize this for the next patch release? The Full Access and missing-prompt failures in #141 still interrupt everyday Codex use in 0.1.42. Please let me know if anything else is needed before merge.

hardbeat920:
Thanks @yankawai I’ve rechecked the latest changes. Both remaining issues are resolved, CI is green. This is ready to merge.

PR #143, opened 2026-09-09
Fix session rename freezing while the agent is working

### LINKED ISSUE #142: [Bug] Rename freezes for working sessions and blocks opening the conversation
### MonoCode version

Reproduced against 0.1.41 source, commit `20b78f5f4471b3303fa6b5c05cd54804b8af1d22`.

### Operating system

macOS 26.5.2

### Provider CLI

Codex. The disabling condition is in the shared sidebar, independent of provider-specific code.

### What happened

Choosing **Rename** on a working conversation replaces its sidebar card with a disabled text field. The field looks like the normal inline editor, but cannot receive focus or text input. Enter and Escape cannot reach the editor's save/cancel handlers. Because the clickable session card has been replaced, the conversation also cannot be opened through that row while it is stuck in rename mode.

This has a serious usability impact during long-running turns: a routine menu action blocks both renaming and normal navigation from the affected row. Users should not need to wait for or stop an agent just to edit a conversation title. A rename started while idle also becomes disabled if the session starts working during the edit.

Expected: the existing inline editor stays usable while the agent works; Enter or blur saves, Escape cancels, and the regular clickable card returns afterward.

### Steps to reproduce

1. Start a long-running turn in a conversation visible in the sidebar.
2. While its status is **Working...**, right-click the conversation and choose **Rename**. F2 on the focused card reaches the same editor.
3. Try editing the title, saving with Enter, or cancelling with Escape.
4. Observe the disabled field in place of the normal card. Clicking that row no longer opens the conversation.

### Cause and proposed fix

The menu allows Rename while a session is busy, but [`SessionRenameRow`](https://github.com/hardbeat920/monocode/blob/20b78f5f4471b3303fa6b5c05cd54804b8af1d22/src/chrome/Sidebar.tsx#L2494-L2573) renders its input with `disabled={busy}`. Disabled inputs cannot receive the focus required by the existing keyboard and blur handlers.

Remove the busy condition and its now-unused prop from this editor. The existing rename callback already updates and persists the session title without a busy restriction. Keep the current component, styling, and save/cancel behavior.

### Reproduction evidence

DOM interaction tests using the real Sidebar and context menu reproduce this on the upstream commit: the idle case passes, while seven working-session/focus/lifecycle cases fail. With the focused fix, all eight pass, including the needs-approval state and normal navigation after rename. Native services are stubbed in these UI tests; this is not a claim of a released desktop fix.

yankawai:
## What changed

Keep the sidebar's inline session rename input editable while the agent is working. Remove the unused `busy` prop from `SessionRenameRow` and add eight DOM interaction regression tests.

Fixes #142.

## Why

The context menu and F2 could start Rename for a working session, but the editor immediately rendered `disabled={busy}`. It could not receive focus, typing, Enter, Escape, or the normal blur-to-save interaction. The disabled editor also replaced the clickable session card, blocking navigation through that row during a long-running turn. Starting a turn during an existing rename disabled it as well.

Title changes already have an independent update/persistence path. Removing the input's busy restriction lets the existing save/cancel handlers finish and restore the card.

## Validation

- On upstream `20b78f5`: the new idle test passes; seven working-session/focus/lifecycle tests fail.
- After the fix: all eight pass. Coverage includes context menu and F2 entry, focus/selection, Enter, Escape, blur, blank names, needs-approval state, navigation after saving, and a turn starting during editing.
- `npm run check`: 141 test files / 1433 frontend tests, TypeScript, Rust formatting, Clippy with warnings denied, and 221 Rust tests pass.
- `npm run build` passes, with Vite's mixed-import and large-chunk warnings.

UI interaction tests mount the actual Sidebar/menu and stub unrelated native services. The installed desktop application has not been replaced or verified end to end with this patch.

## UI

The existing inline editor, menu, classes, spacing, colors, and keyboard/mouse behavior are retained. The production change removes four lines; there is no layout or component redesign.

## Checklist

- [x] I ran `npm run check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Session names can now be edited while a session is busy or a turn is in progress.
  - Renaming remains available across different session states, including working and approval-required states.

- **Tests**
  - Added coverage for context-menu and keyboard-based renaming, cancellation, commit-on-blur, empty names, and editing during an active turn.

### DIFF
--- a/src/chrome/Sidebar.tsx
+++ b/src/chrome/Sidebar.tsx
@@ -903,7 +903,6 @@ function SidebarComponent({
       <SessionRenameRow
         session={session}
         isActive={session.id === activeSessionId}
-        busy={busySessionIds.has(session.id)}
         needsApproval={approvalSessionIds.has(session.id)}
         onCommit={(title) => {
           onRenameSession(session.id, title);
@@ -2494,14 +2493,12 @@ function SessionCard({
 function SessionRenameRow({
   session,
   isActive,
-  busy,
   needsApproval,
   onCommit,
   onCancel,
 }: {
   session: SessionSummary;
   isActive: boolean;
-  busy: boolean;
   needsApproval: boolean;
   onCommit: (title: string) => void;
   onCancel: () => void;
@@ -2560,7 +2557,6 @@ function SessionRenameRow({
       <input
         ref={inputRef}
         value={value}
-        disabled={busy}
         onChange={(e) => setValue(e.target.value)}
         onBlur={() => finish(true)}
         onKeyDown={onKeyDown}
--- a/src/chrome/SidebarRename.test.ts
+++ b/src/chrome/SidebarRename.test.ts
@@ -1,0 +1,226 @@
+// @vitest-environment happy-dom
+import { act, createElement, type ComponentProps } from "react";
+import { createRoot, type Root } from "react-dom/client";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { formatSessionTitle } from "../lib/session";
+import { Sidebar } from "./Sidebar";
+
+// Keep native services out of these menu/input interaction tests.
+vi.mock("../hooks/useInboxUnseen", () => ({ useInboxUnseen: () => false }));
+vi.mock("../hooks/useProjectDiffStats", () => ({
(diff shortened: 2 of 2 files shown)


hardbeat920:
@yankawai nice catch

PR #146, opened 2026-09-09
Fix Markdown file links and source navigation

yankawai:
## What changed

Make local file references in the conversation open the existing right-hand editor at their line and column. Normalize Markdown file URLs before sanitization, preserve source positions, and show Markdown source when a link targets a line. The existing contents, diff views and file-link context menu remain available.

## Why

Several parts of the same file-opening flow failed:

- Protocol methods such as `currentTime/read` appeared as file chips, while extensionless files such as `Dockerfile` were not openable.
- Relative Markdown links and `file://` links could be blocked before reaching the file handler. Encoded filenames and source locations could be lost. Heading anchors are separated from filenames before decoding, preserving literal encoded hash characters.
- An optional project-index failure prevented even a direct file path from opening.
- Bare references such as `main.ts:12` were mistaken for URL schemes. Source locations are removed before scheme validation.
- Encoded network paths, including `file://localhost/%2Fhost/share/file.md` and mixed Windows separators, could reach the native reader. Validate after decoding and separator normalization for both link and native-path resolution.
- A location beyond the current end of a file retried indefinitely. Clamp it immediately and retry after reload only while its line is still pending. Consume completed navigation and cancel pending retries when the user changes selection, edits, or leaves the editor, so later reloads cannot jump back or steal focus. New explicit navigation requests still work.

The URL sanitizer and external-link behavior remain in place. A failed optional index lookup falls back to the editor's normal file read and error display.

## UI and compatibility

Uses the existing file chips, right-hand pane, source/preview controls, diff rendering and upstream file-link context menu. No layout, CSS or theme changes. “Open in MonoCode” preserves line/column information; copy, reveal and external-open actions receive the file path without a source location.

Component tests exercise the real Markdown renderer, context menu and file pane. Native filesystem/git calls are mocked in those tests; the installed application was not replaced.

## Validation

- `npm run check` on the changes integrated with main `68319d8`: **1,686 web tests and 230 Rust tests pass**, along with TypeScript, Rust formatting and Clippy.
- `npm run build`: passes.
- The original regressions reproduced 15 failures before the earlier fixes. Four additional cases reproduce completed-navigation replay, cursor movement and focus loss on the previous PR head, and pass with this update.
- 85 focused cases cover path resolution, rendered links, context-menu actions and editor navigation, including delayed content, cancellation and a new explicit request.
- Existing tests cover extensionless files, encoded filename characters, Windows paths, local file URLs, blocked network paths and non-file schemes, and the existing added/removed-line diff.

The CI workflow and Linux dependency setup have no changes relative to current upstream.

## Checklist

- [x] I ran `npm run check`
- [x] This PR is focused on file links and source navigation

## Summary by CodeRabbit

* **New Features**
  * Markdown links to workspace files open the correct file, including line and column locations.
  * Supports local paths, file URLs, extensionless files, anchors, and Windows-style paths.
  * File navigation automatically switches to source view and highlights the requested location.
  * Adds actions for working with selected text in the file editor.

* **Bug Fixes**
  * File opening continues to work when the workspace file index is unavailable.
  * Navigation remains accurate after files reload or change in size.
  * External, unsafe, and network file links are handled safely.


hardbeat920:
Hey @yankawai, please remove the CI and Linux dependency setup changes from this PR. I’d like to keep the current CI configuration and keep this focused on the file-link and navigation fixes.

Also the approach looks sensible, but I noticed the new parser rejects bare file references like main.ts:12 as URL schemes. Encoded paths like/%2Fhost/share/file.md also bypass the network-path check after decoding. 

Could you address both and add regression tests before we merge?

hardbeat920:
Hey @yankawai, thanks for addressing the feedback! One remaining issue, after opening a link and moving to another line, a file reload jumps back to the original line and grabs focus again. 

Could you stop replaying the navigation once it’s completed or the user has moved elsewhere, and add a regression test?

yankawai:
@hardbeat920 addressed the reload issue: completed navigation is consumed once, and a pending clamped target is cancelled when the user moves the selection, edits the file, or leaves the editor. Reloads preserve the current caret and focus. A new explicit link request still works, and a pending target can still resolve when its line arrives.

Four regression cases fail on the previous PR head and pass with the fix. I also integrated the current main and resolved the conflict with the new file-link context menu, preserving line navigation through “Open in MonoCode”.

Validation on the combined changes: `npm run check` passes (1,686 web tests, 230 Rust tests, TypeScript, formatting and Clippy), and `npm run build` passes. The existing UI styling is unchanged. This is ready for another review; please consider it for the next patch release.

hardbeat920:
@yankawai thank you this is great. merging

PR #155, opened 2026-09-10
Fix OMP advisor interjection rendering

### LINKED ISSUE #123: OMP: advisor interjections are invisible and their extra phases collapse into the single transcript fold, making a long answer read as "folded into thinking" with only a short tail shown
### Summary

When OMP's **advisor** interjects mid-turn (OMP injects a `custom_message` and the agent runs an extra continuation phase), MonoCode gives that interjection **zero rendering**, and the whole affected work region then collapses into the **single transcript fold**. What the user perceives: their question → one "Worked for…" line → a **short tail answer**, with everything substantive (including complete earlier answer prose and the advisor-triggered reasoning) hidden behind thinking rows. It reads exactly like "MonoCode misjudged my answer as thinking and folded it", but the block roles are actually correct — the content disappears into the fold because the interjection boundaries are invisible.

### Environment

- MonoCode 0.1.38 (observed); verified against `main` @ `d4cd7df` — code paths unchanged
- OMP `@oh-my-pi/pi-coding-agent` 18.1.14, advisor enabled (`advisor.enabled: true`, advisor model gemini-3.8-flash)
- macOS 26.x (Apple Silicon), harness session: OMP over `--mode rpc`

### Real-world trace (OMP session, 2026-09-08)

One user prompt, one logical "answer", but the advisor fired mid-turn twice. OMP session entries (UTC):

```
13:32:52 user      "对嵌入的这种视频图片，你都用OSS链接…"
13:33:04 assistant thinking(1192 chars) + toolCall
13:33:04 toolResult (curl failed)
13:33:04 custom_message customType=advisor display=true   ← interjection #1
13:33:17 assistant thinking(1085) + toolCall              ← agent pivots on the advisory
… tools …
13:33:59 assistant thinking(115) + toolCall
13:34:09 assistant text(644)                              ← final answer, the ONLY prose
```

An earlier exchange the same hour shows the same shape with a twist:

```
13:30:06 assistant text(936)  "完成。交付清单：…"          ← complete answer
13:30:06 custom_message advisor                           ← interjection #2
13:30:26 assistant thinking(3400) + tools …               ← extra work phase
13:31:31 assistant text(683)                              ← second "answer"
```

### What MonoCode renders

Replaying these frames through MonoCode's own pipeline (`piFamily.handleFrame` → `applyHarnessEvent` → `groupTurnItems`/`foldableWork`, verified in a vitest run against `main`):

- **Visible at full strength**: the user message + the 644-char tail answer. In the earlier exchange: two prose blocks (936 and 683 chars) with a fold line between them — two "answers" for one prompt, no visible cause.
- **Behind the single "Worked for…" fold**: ~2,400 chars of reasoning rows (including the advisory-weighing thinking) + all tool rows.
- **Invisible entirely**: both advisor `custom_message` entries. The user has no way to know the agent was interrupted and re-scoped.

So a long, complete reply appears as "a small output with everything folded into thinking" — and the thinking rows do contain prose-like substance (the continuation reasoning discusses the advisory in full sentences), which completes the illusion.

### Root cause / locations

1. **`src/lib/harness/piFamily.ts` — `handleFrame()`**: OMP-specific frame handling covers `session_info_update`, `config_update`, `command_output`, `prompt_result`, async `prompt` errors, and `agent_end.isTerminal`. There is **no case for**:
   - `custom_message` (advisor interjections; OMP sends `display: true` ones exactly so clients can show them) → dropped silently;
   - `advisor_yielded` → dropped (would be a natural status-line chip: "advisor reviewed this turn");
   - `message_start` / `turn_start` / `turn_end` → dropped, so a mid-turn interjection produces no visual boundary between the pre-advisory and post-advisory segments.
2. **`src/surfaces/transcriptActivity.ts` — `foldableWork()`**: folds the last activity group before the final prose. Because the advisor's extra phases extend that group *after* a complete answer was already streamed, the earlier full-strength answer ends up sandwiched above a fold line, and the only prominent text left is the short post-advisory ack. (Parser roles themselves are correct — a frame-replay test yields `reasoning / assistant / reasoning / assistant` as expected. Nothing is mislabeled; it just renders as if it were.)

### Reproduction

1. OMP with `advisor.enabled: true` (any model; a flash-tier advisor interjects often).
2. In MonoCode, give OMP a multi-step task with tool calls.
3. Watch the RPC stdout: between `turn_start` frames you'll see `{"type":"advisor_yielded"}` and (when the advisor has guidance) a `custom_message` frame.
4. MonoCode: no divider/chip/status appears; after settle, everything since the prompt collapses behind one "Worked for…" line and only the short post-advisory text stands alone. Expanding thinking rows shows the advisory discussion — the report I received from a user was literally "MonoCode misjudged everything above as thinking because an advisor interjected".

Minimal frame replay (mocked child, real frame shapes from an 18.1.14 capture):

```ts
frame(sid, delta("thinking", "…")); frame(sid, delta("text", "full answer", 1));
frame(sid, { type: "agent_end", isTerminal: false, messages: [] });
frame(sid, { type: "custom_message", customType: "advisor", content: "<advisory>…</advisory>", display: true, attribution: "agent" });
frame(sid, { type: "advisor_yielded" });
frame(sid, { type: "turn_start" });
frame(sid, delta("thinking", "reacting to advisories")); frame(sid, delta("text", "Short ack.", 1));
frame(sid, { type: "agent_end", isTerminal: true, messages: [] });
// blocks come out as [reasoning, assistant, reasoning, assistant] — roles fine,
// but groupTurnItems/foldableWork render: fold(2 thinking) + "Short ack." only
```

### Suggestions

- Render `custom_message` with `display: true` (at minimum `customType: "advisor"`) as a visible interjection divider/chip in the transcript, and surface `advisor_yielded` as a transient status line.
- Use `turn_start`/`message_start` boundaries to seal the current text segment, so a pre-advisory answer doesn't visually merge into the advisor-steered continuation.
- Optionally: when an interjection lands after a prose block, don't let the continuation extend the *same* fold group the earlier answer already "closed" — or auto-expand that group once.

Happy to share the full captured frame log or the session JSONL excerpt if useful.

elijah7x:
## What changed

- Handle OMP `message_start` / `message_end` pairs whose message has `role: "custom"`.
- Render displayed custom messages as persisted transcript dividers. Advisor cards use structured note text and the highest reported severity.
- Keep hidden custom messages hidden, preserve Pi behavior, and retain interjection metadata across restarts.
- Treat the divider as a fold boundary so a complete pre-advisor answer stays visible.

## Why

OMP emits advisor cards over RPC as `message_start` followed by `message_end`; the nested message carries `role: "custom"`, `customType`, `display`, `content`, and `details`. MonoCode ignored that message role. Without a visible boundary, the post-advisor work phase caused `foldableWork()` to fold the complete answer that preceded it.

Issue #123 calls the live frame `custom_message`, which is the session JSONL entry type. This patch follows the wire shape from OMP 18.1.16 and emits one divider on `message_start`; it consumes `message_end` without duplicating the row.

Closes #123.

## UI

Before: the fold contains the first work phase, the complete answer, and the post-advisor work. The transcript shows only the prompt, one work row, and the short tail reply.

After: the complete answer stays visible above an `Advisor` divider. The divider shows the structured note and severity; only the work after the advisor folds under the later work row.

I rendered both states at 1598 x 870 through the real `AgentTranscript` component and checked the screenshots for clipping, overflow, hierarchy, and fold placement.

## Checklist

- [x] I ran `npm run check`

## Summary by CodeRabbit

- **New Features**
  - Added support for displaying advisor and other visible system interjections in the transcript.
  - Interjections can show custom labels, message content, and severity levels.
  - Interjections persist across session restarts and visually separate surrounding work phases.
  - Hidden internal messages and unsupported content are excluded from the transcript.

- **Bug Fixes**
  - Prevented transcript folding from obscuring work completed before an interjection.
  - Added validation to discard malformed interjection metadata while preserving the system message.

### DIFF
--- a/src/lib/harness/apply.test.ts
+++ b/src/lib/harness/apply.test.ts
@@ -161,6 +161,55 @@ describe("status blocks", () => {
   });
 });
 
+describe("interjection blocks", () => {
+  it("seals assistant streams on both sides of the boundary", () => {
+    let session = appendUser(newSession("pi", "/tmp"), "go");
+    session = applyHarnessEvent(session, {
+      type: "message.delta",
+      text: "Complete answer.",
+    });
+    session = applyHarnessEvent(session, {
+      type: "interjection",
+      text: "Check the fallback.",
+      customType: "advisor",
+    });
+    session = applyHarnessEvent(session, {
+      type: "message.delta",
+      text: "Checked.",
+    });
+
+    expect(session.blocks.slice(1)).toMatchObject([
+      { role: "assistant", text: "Complete answer.", streaming: false },
+      { role: "system", interjection: { customType: "advisor" } },
+      { role: "assistant", text: "Checked.", streaming: true },
+    ]);
+  });
+
+  it("appends every interjection as a distinct persisted boundary", () => {
+    let session = appendUser(newSession("pi", "/tmp"), "go");
+    session = applyHarnessEvent(session, {
+      type: "interjection",
+      text: "Check the fallback.",
+      customType: "advisor",
+      severity: "concern",
+    });
+    session = applyHarnessEvent(session, {
+      type: "interjection",
+      text: "Check the fallback.",
+      customType: "advisor",
+      severity: "concern",
+    });
+
+    const interjections = session.blocks.filter((block) => block.interjection);
+    expect(interjections).toHaveLength(2);
+    expect(interjections[0]).toMatchObject({
+      role: "system",
+      text: "Check the fallback.",
+      interjection: { customType: "advisor", severity: "concern" },
+    });
+  });
+});
+
 describe("task list updates", () => {
   it("updates one structured checklist instead of appending plan cards", () => {
     let session = appendUser(newSession("codex", "/tmp"), "fix it");
@@ -454,17 +503,26 @@ describe("applyHarnessEvent context", () => {
 describe("tool enrichment", () => {
   it("retains Edit and Write previews when a tool completes without repeating its input", () => {
     for (const [name, input] of [
-      ["Edit", { file_path: "/notes.md", old_string: "old", new_string: "new" }],
+      [
+        "Edit",
+        { file_path: "/notes.md", old_string: "old", new_string: "new" },
+      ],
       ["Write", { file_path: "/notes.md", content: "  content\n" }],
       ["Write", { file_path: "/notes.md", content: "" }],
     ] as const) {
       const preview = previewFromTool(name, input)!;
       let session = applyHarnessEvent(newSession("claude", "/repo"), {
-        type: "tool.started", callId: "edit", title: name, kind: "edit",
-        status: "pending", preview,
+        type: "tool.started",
+        callId: "edit",
+        title: name,
+        kind: "edit",
+        status: "pending",
+        preview,
       });
       session = applyHarnessEvent(session, {
-        type: "tool.updated", callId: "edit", status: "completed",
+        type: "tool.updated",
+        callId: "edit",
+        status: "completed",
       });
       expect(session.blocks[0].tool?.preview).toMatchObject(preview);
       expect(session.blocks[0].tool?.preview?.lines).toEqual(preview.lines);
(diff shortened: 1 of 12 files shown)


elijah7x:
Withdrawing this for now: validation against a real persisted OMP session shows the advisor interjection still remains inside the same collapsed work group, so the user-visible bug is not fixed yet. I’ll continue debugging locally and only reopen or submit a replacement after the real-session behavior is verified.

PR #156, opened 2026-09-10
Fix OMP advisor interjections folding away complete answers

### LINKED ISSUE #123: OMP: advisor interjections are invisible and their extra phases collapse into the single transcript fold, making a long answer read as "folded into thinking" with only a short tail shown

elijah7x:
# Fix OMP advisor interjections folding away complete answers (#123)

Fixes #123

## Problem

When OMP's advisor interjects mid-turn, OMP emits `custom_message` (customType `advisor`, `display: true`) and `advisor_yielded` frames on the RPC stream. MonoCode dropped them silently, which had two visible effects:

- **Live turns**: the interjection was invisible, and worse, the fold logic could no longer distinguish "progress commentary" from "a complete answer followed by a new segment" — the complete earlier answer folded behind the work chrome, leaving only a short tail visible (screenshots in #123).
- **Already-saved sessions**: builds before this fix persisted `assistant → reasoning/tool → assistant` without the boundary. Some even **coalesced the two answers into a single assistant block** (streaming text appended across the boundary). Re-rendering those sessions still folded the complete answer.

## Fix, part 1 — live rendering

- `piFamily.ts` (omp flavor): `handleFrame` emits a new persisted `interjection` harness event for `custom_message` frames with `display: true` (structured advisor notes with severity `nit|concern|blocker`, plus generic custom types); `advisor_yielded` becomes a transient status line. `display: false` stays hidden.
- `apply.ts`: the event appends a `system` block carrying `interjection` chrome (`customType`, max `severity`).
- `AgentTranscript.tsx`: renders a distinct interjection row (severity styled) instead of an invisible system block.
- `transcriptActivity.ts`: interjection blocks are neither prose nor work, so `foldableWork` cannot extend a fold across them — an answer the harness already showed never folds behind a later interjection.

## Fix, part 2 — backfill for sessions saved by older builds

- New Tauri command `omp_session_interjections(providerSessionId)` (`src-tauri/src/fs.rs`) scans OMP's own session JSONL (`~/.omp/agent/sessions/**/*_<providerSessionId>.jsonl`) and returns display-true anchors: interjection text/type/severity, the **full text** of the preceding assistant message, its 1-based occurrence among identical texts, and — when the very next message is a text-only assistant reply — its full text as split evidence.
- `sessionStore.getSession` repairs persisted OMP sessions (only when harness is omp and a provider session id is bound): inserts the missing boundary blocks and, where an old build coalesced two answers, splits the block back into `assistant → interjection → assistant`.
- Matching is deliberately strict: exact full-text equality + occurrence disambiguation + source order; a coalesced block is split only when both source texts concatenate exactly to the persisted text. Anything that cannot be matched exactly is left untouched — the repair never guesses boundaries (progress prose between tool calls must keep folding).
- Deterministic block ids (`omp-interjection-<omp message id>`) make the repair idempotent; repaired sessions are persisted once via the normal upsert path. Other harnesses and unbound sessions are untouched; any source/log error degrades silently to today's rendering.

## Verification

- `npx vitest run`: 1515 tests green (1505 baseline + 10 new covering: fold no longer swallows the earlier answer, occurrence disambiguation, no-anchor turns unchanged, no fuzzy/prefix matching, idempotent re-run, live-random-id dedup, coalesced-block split, load persistence, error degradation).
- `src-tauri`: `cargo fmt --check` + `cargo check` clean; 2 new parser tests (anchor gating, no coalesce-evidence across thinking parts).
- Real-data validation on the reporter's machine: 13 affected persisted sessions; the heaviest (17 anchors in one session) matched 17/17 — 15 boundary insertions plus 2 exact coalesced-block splits (e.g. 1146+401=1547 chars restored to two answers); block count 1353 → 1372; second pass changed nothing.

## Notes

- Only the OMP flavor is affected; pi/fx/claude/codex/... paths are unchanged.
- No schema migration: repair happens lazily at load from OMP's own logs, so sessions saved by even older builds heal themselves without a one-shot migration pass.

## Summary by CodeRabbit

- **New Features**
  - Added advisor notes and custom messages as transcript interjections with severity levels.
  - Added status updates when an advisor reviews a turn.
  - Recovered missing interjections from existing OMP session transcripts.
  - Interjections remain visible in session history and mark work-phase boundaries.
  - Long interjection messages can be expanded or collapsed.

- **Bug Fixes**
  - Improved matching and counting of repeated transcript content during recovery.
  - Preserved streaming content across status and interjection rows.
  - Preserved transcript content when recovery or persistence encounters errors.
  - Session setting updates now retain existing values not included in updates.

### DIFF
--- a/src-tauri/src/fs.rs
+++ b/src-tauri/src/fs.rs
@@ -47,6 +47,24 @@ pub struct OmpInterjectionAnchor {
 pub fn omp_session_interjections(
     provider_session_id: String,
 ) -> Result<Vec<OmpInterjectionAnchor>, String> {
+    let Some(path) = omp_session_path(&provider_session_id)? else {
+        return Ok(Vec::new());
+    };
+    parse_omp_interjections(&path)
+}
+
+#[tauri::command(async)]
+pub fn omp_verify_assistant_texts(
+    provider_session_id: String,
+    texts: Vec<String>,
+) -> Result<Vec<bool>, String> {
+    let Some(path) = omp_session_path(&provider_session_id)? else {
+        return Ok(vec![false; texts.len()]);
+    };
+    verify_omp_assistant_texts(&path, &texts)
+}
+
+fn omp_session_path(provider_session_id: &str) -> Result<Option<PathBuf>, String> {
     if provider_session_id.is_empty()
         || !provider_session_id
             .bytes()
@@ -58,10 +76,7 @@ pub fn omp_session_interjections(
         .map(PathBuf::from)
         .ok_or("Home directory is unavailable")?
         .join(".omp/agent/sessions");
-    let Some(path) = find_omp_session_file(&root, &provider_session_id) else {
-        return Ok(Vec::new());
-    };
-    parse_omp_interjections(&path)
+    Ok(find_omp_session_file(&root, provider_session_id))
 }
 
 fn find_omp_session_file(root: &Path, provider_session_id: &str) -> Option<PathBuf> {
@@ -115,13 +130,8 @@ fn omp_message_text(value: &serde_json::Value, separator: &str) -> String {
         .join(separator)
 }
 
-fn parse_omp_interjections(path: &Path) -> Result<Vec<OmpInterjectionAnchor>, String> {
+fn read_omp_entries(path: &Path) -> Result<Vec<serde_json::Value>, String> {
     let file = std::fs::File::open(path).map_err(|error| format!("{}: {error}", path.display()))?;
-    let mut assistants: HashMap<&str, (String, usize, String, usize)> = HashMap::new();
-    let mut occurrences: HashMap<String, usize> = HashMap::new();
-    let mut concat_occurrences: HashMap<String, usize> = HashMap::new();
-    let mut out: Vec<OmpInterjectionAnchor> = Vec::new();
-
     let mut entries = Vec::new();
     for line in BufReader::new(file).lines() {
         let Ok(line) = line else { continue };
@@ -130,6 +140,10 @@ fn parse_omp_interjections(path: &Path) -> Result<Vec<OmpInterjectionAnchor>, St
         };
         entries.push(value);
     }
+    Ok(entries)
+}
+
+fn omp_active_ids(entries: &[serde_json::Value]) -> HashSet<&str> {
     let nodes: HashMap<_, _> = entries
         .iter()
         .filter_map(|value| value["id"].as_str().map(|id| (id, value)))
@@ -142,6 +156,43 @@ fn parse_omp_interjections(path: &Path) -> Result<Vec<OmpInterjectionAnchor>, St
         }
         cursor = nodes.get(id).and_then(|value| value["parentId"].as_str());
     }
+    active
+}
+
+fn verify_omp_assistant_texts(path: &Path, texts: &[String]) -> Result<Vec<bool>, String> {
+    let entries = read_omp_entries(path)?;
+    let active = omp_active_ids(&entries);
+    let mut pending: HashSet<&str> = texts.iter().map(String::as_str).collect();
+    for value in &entries {
+        if pending.is_empty() {
+            break;
+        }
+        if value["type"] == "message"
+            && value["message"]["role"] == "assistant"
+            && value["id"].as_str().is_some_and(|id| active.contains(id))
+        {
+            let content = &value["message"]["content"];
+            pending.remove(omp_message_text(content, "\n").as_str());
+            pending.remove(omp_message_text(content, "").as_str());
+        }
+    }
+    Ok(texts
+        .iter()
+        .map(|text| !pending.contains(text.as_str()))
+        .collect())
+}
+
+fn parse_omp_interjections(path: &Path) -> Result<Vec<OmpInterjectionAnchor>, String> {
+    let entries = read_omp_entries(path)?;
+    let nodes: HashMap<_, _> = entries
+        .iter()
+        .filter_map(|value| value["id"].as_str().map(|id| (id, value)))
+        .collect();
+    let active = omp_active_ids(&entries);
+    let mut assistants: HashMap<&str, (String, usize, String, usize)> = HashMap::new();
+    let mut occurrences: HashMap<String, usize> = HashMap::new();
+    let mut concat_occurrences: HashMap<String, usize> = HashMap::new();
+    let mut out: Vec<OmpInterjectionAnchor> = Vec::new();
     // Metadata and compaction participate in ancestry, not anchor text.
     // Keep file order for occurrences and notes, but exclude abandoned branches.
     for value in &entries {
@@ -4194,6 +4245,38 @@ mod tests {
     use std::sync::atomic::{AtomicU64, Ordering};
 
     static TMP_SEQ: AtomicU64 = AtomicU64::new(0);
+
+    #[test]
+    fn omp_assistant_verification_requires_exact_active_message_text() {
+        let dir = tmp("omp-verification");
+        let path = dir.0.join("session.jsonl");
+        let records = [
+            serde_json::json!({"type":"message","id":"u","message":{"role":"user","content":"User only"}}),
+            serde_json::json!({"type":"message","id":"abandoned","parentId":"u","message":{"role":"assistant","content":"Off branch"}}),
+            serde_json::json!({"type":"message","id":"a","parentId":"u","message":{"role":"assistant","content":[{"type":"text","text":"First."},{"type":"thinking","thinking":"Hidden"},{"type":"text","text":"Second."}]}}),
+            serde_json::json!({"type":"message","id":"b","parentId":"a","message":{"role":"assistant","content":"Plain"}}),
+            serde_json::json!({"type":"custom_message","id":"note","parentId":"b","content":"Note only"}),
+        ];
+        let jsonl = records.iter().map(|v| format!("{v}\n")).collect::<String>();
+        std::fs::write(&path, jsonl).unwrap();
+        let texts = [
+            "First.\nSecond.",
+            "First.Second.",
+            "Plain",
+            "Off branch",
+            "User only",
+            "Note only",
+            "First.",
+            " First.Second.",
+            "First. Second.",
+            "First.Second.",
+        ]
+        .map(str::to_owned);
+        assert_eq!(
+            verify_omp_assistant_texts(&path, &texts).unwrap(),
+            [true, true, true, false, false, false, false, false, false, true]
+        );
+    }
 
     #[test]
     fn omp_interjections_require_displayed_assistant_anchors() {
--- a/src-tauri/src/fs.rs
+++ b/src-tauri/src/fs.rs
@@ -1,6 +1,5 @@
 use std::collections::{HashMap, HashSet};
-use std::io::ErrorKind;
-use std::io::Write;
+use std::io::{BufRead, BufReader, ErrorKind, Write};
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 use std::sync::Mutex;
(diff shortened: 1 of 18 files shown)


hardbeat920:
Thanks for the thorough fix, @elijah7x! 

The live rendering looks good, but I found a few concerns in the persisted-session repair:

- It scans every JSONL entry instead of only the active OMP branch, so interjections from abandoned branches could be restored and persisted incorrectly.
- Interjections whose immediate parent is a tool result or metadata entry are skipped, including the toolResult > custom_message sequence from #123.
- omp_message_text joins multiple text parts with newlines, while MonoCode’s live stream concatenates them directly, which can prevent exact matching.

Could you take another look at these cases and add regression tests before we merge?

elijah7x:
@hardbeat920 yep working on it

elijah7x:
Thanks for the careful review — all three concerns are addressed in bda9880, plus one related live-streaming fix in 4156961:

- **Active branch**: the parser now indexes records by id and reconstructs the active path by walking `parentId` links back from the last entry; only on-path records participate in anchor construction and occurrence counting. Regression-tested with a synthetic abandoned branch. (For what it's worth, we found no abandoned-message forks in the reporter's real logs — treating this as defense-in-depth.)
- **toolResult / metadata parents**: confirmed substantive — in the reporter's data 67 of 126 `display:true` interjections are parented to `toolResult` messages. The backfill now resolves non-assistant parents by walking the parent chain to the nearest assistant ancestor and anchoring after that prose (conservatively skipping user / dead-end / blank-prose chains; multiple notes on the same anchor stack in source order).
- **Text-part joining**: real logs show zero multi-part assistant contents so far, but the matcher now also accepts the direct-concatenation form (exact equality only, per-form occurrence counting), with a synthetic regression.
- **4156961**: while testing, the reporter hit a related live issue — an advisor status row arriving mid-stream split the still-streaming assistant message into two persisted blocks (status wedged mid-sentence). Streaming now continues through status/interjection rows (a delta after a *completed* message still opens a new block), and the persisted-session repair conservatively merges already-split pairs when they exactly match a single JSONL message.

Gates: vitest 1529 tests green (148 files), `cargo fmt --check` / `cargo check` / parser tests clean, `tsc` clean. End-to-end validated against a copy of the reporter's real DB using the shipped Rust parser + TS matcher: restored boundaries in the heaviest session went 17 → 32 with previous ones unchanged and the repair idempotent; the split message merged back into one block with a later anchor landing correctly. For transparency: 65 of the 99 `display:true` entries in that session stop at a textless assistant (nothing to seal) and are skipped by design, and 2 anchors remain unmatched under strict-equality matching.

elijah7x:
Follow-up (add1907): while testing 4156961 we found one more gap in the status-split repair — a bare `advisor_yielded` (advisor reviewed the turn but produced no note) also split a streaming message, and such splits had no adjacent anchor text to serve as merge evidence. The repair now collects status-split candidates and verifies each `A+B` concatenation against the full set of active-path assistant message texts via a second command (`omp_verify_assistant_texts`) — exact equality in both text-part forms, only verified candidates merge, and the anchor-based repair path is unchanged. Validated end-to-end against the reporter's real DB: the previously-stuck split now merges with zero adjacent anchors (status row preserved after the merged block), session-level repair is idempotent, and the heaviest session is unchanged (32 boundaries / 2057 blocks). Gates: vitest 1532 tests green (148 files), `cargo fmt --check` / `cargo check` / 235 Rust tests, `tsc` clean.

hardbeat920:
Thanks for the follow-up, @elijah7x. add1907 fixes the unanchored advisor_yielded case, but I think two concerns remain:

- Interjections are still skipped like ordinary status rows during streaming, so separate assistant messages around an interjection can be merged.
- Status-split verification checks whether A+B matches any active assistant message, but it is not tied to the specific occurrence. An unrelated message with the same text could therefore authorize an incorrect merge.

Could you make interjections hard stream boundaries and make the persisted repair occurrence-aware, with regression tests for both cases?

elijah7x:
Thanks — both remaining cases should be closed in 14c5fbf / b339a78.

Interjections are hard stream boundaries. `patchStreaming` / `sealLastStream` skip only ordinary status rows (system without interjection meta), matching `statusSplitEnd`. Appending an interjection seals the open stream first, so a later delta cannot glue two assistant messages across the divider. `advisor_yielded` status still continues an in-flight message. Completed message → interjection → new message stays two blocks, same as #123.

Status-split repair is occurrence-aware. `omp_verify_assistant_texts` now returns per-text occurrence bounds on the active path instead of a boolean set. The persisted walk authorizes a merge only for `seen[T]+1` when that occurrence exists, and consumes it — including anchor-derived evidence. Two identical splits with one source message repair only the first; a complete `T` before a later split occupies occurrence 1.

Regressions cover both, plus stacked interjections, status after an interjection, and an occupancy cap (10 identical splits, bound 3). Happy to squash if you'd rather keep the PR to one commit.

hardbeat920:
Thanks, @elijah7x the hard stream-boundary change looks good and addresses the first concern.

I think the persisted repair still has an ordering collision. 

The occurrence count says how many source messages equal T, but not where they occur. For persisted blocks [A, status, B, AB] with source messages [A, B, AB], the first A+B candidate is treated as occurrence 1 of AB and merged, even though the real AB is the later complete block. The new tests cover a complete AB before the split, but not this reverse ordering.

Let me know if this makes sense and if so could you add that regression and preserve source position/order in the verification?

elijah7x:
That's reasonable. I'll make some changes.

hardbeat920:
Thanks, @elijah7x the ordered alignment fixes my previous concern. 

One small safety issue remains, when the alignment exceeds its budget, it falls back to the weaker linear matcher, which can reintroduce an incorrect merge. 

Since the repair is persisted, could this path fail closed and return the original blocks instead?

Thanks again for everything 🙏

elijah7x:
Agreed — done in 43fba6b. When the alignment throws on its budget (depth / memo / concatenation), `mergeStatusSplits` now returns the original blocks untouched; `mergeStatusSplitsLinear` is removed entirely, so there is no weaker matcher left to reintroduce a weld. Because the same reference flows out of `backfillOmpInterjections`, `getSession` also skips the persist write on that path.

Regression: a 2,500-fragment status-split chain whose full concatenation exists in the source trips the depth budget and comes back unmerged (same reference), with a 3-fragment control of the same shape confirming the normal path still welds. Web gates: vitest 185 files / 2097 tests, `tsc` clean.

elijah7x:
Also merged current main (8d3efad, one trivial import conflict in fs.rs) — the PR is mergeable again. Full gates green: vitest 186 files / 2113 tests, cargo test 259, fmt / clippy / tsc clean. Should be ready to go whenever you are.

hardbeat920:
@elijah7x thank you for this 🙏 Merged

PR #177, opened 2026-09-11
Resolve shortened transcript file paths

yankawai:
## What changed

Shortened file links in a transcript resolve against the exact paths that turn's structured tool events already carry. Indexed project files keep precedence, an ambiguous match stays unresolved rather than guessed, and the link's context menu resolves the same way a click does.

## Why

When a session starts in one directory and the agent edits a file in another project, a later `backup.yaml` chip resolved to `<session cwd>/backup.yaml` and the editor showed ENOENT, although the transcript already held the exact path.

Three things came out of the review, all of them here:

- the open-file callback no longer rebuilds per streamed block. The transcript holds `blocks` and `cwd` in a ref and collects candidate paths when a link is clicked, so `sameActivity` keeps memoising settled groups through a long turn;
- `tool.started` and `tool.updated` carry every path of a multi-file edit onto the row, so each link in the answer resolves rather than only the previewed file, and a later progress patch no longer drops the list;
- Reveal, Open in Default App and both copy actions ask for the same resolution a click gets, instead of acting on the text of the link.

Main is merged in rather than rebased, so the review above stays attached to the commit it was written against.

## UI

No layout change. Before, a shortened chip could open a nonexistent path under the session cwd, and Reveal pointed at the same missing file. After, both land on the file the transcript recorded.

## Checklist

- [x] I ran `npm run check`

`tsc --noEmit` is clean. The PR adds ten cases across the reducer, the path collector, transcript links and the link menu; reverting the fixes fails four of them, which is the control for the three points above. A full `vitest` run matches an unmodified `main` on this machine — 23 files and 188 tests fail there too, since Node 26 wants `--localstorage-file` — and `check:rust` stops on `clippy::nonminimal_bool` at `src-tauri/src/control.rs:290` on `main` as well. This PR touches no Rust.

## Summary by CodeRabbit

- **New Features**
  - File links in transcripts now resolve more reliably, including shortened paths and files referenced by tools.
  - Opening, revealing, and copying file links uses the resolved workspace path when available.
  - Multi-file edits retain all associated file paths for easier navigation.

- **Bug Fixes**
  - Improved handling of ambiguous or missing project index matches by falling back to known transcript paths.

- **Tests**
  - Added coverage for multi-file edits, path deduplication, link resolution, and file-opening behavior.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3527,10 +3527,14 @@ export default function App({
   }, []);
 
   const onOpenFile = useCallback<OpenFileFn>(
-    (path, navigation) => {
+    (path, navigation, hints) => {
       void (async () => {
         const resolved =
-          (await resolveOpenablePath(gitCwdRef.current, path)) ?? path;
+          (await resolveOpenablePath(
+            gitCwdRef.current,
+            path,
+            hints?.candidatePaths,
+          )) ?? path;
         rememberOpenedFile(sidebarCwdRef.current, resolved);
         const tab = tabsRef.current.find((entry) => entry.id === activeTabId);
         if (!tab) return;
--- a/src/lib/fileIndex.test.ts
+++ b/src/lib/fileIndex.test.ts
@@ -77,6 +77,27 @@ describe("resolveOpenablePath", () => {
     expect(resolved).toBe(files[2].path);
   });
 
+  it("uses a unique tool path when a short link is outside the project cwd", async () => {
+    const path = "/Users/me/other/project/platform/backup.yaml";
+    const resolved = await resolveOpenablePath(cwd, "backup.yaml", [path]);
+    expect(resolved).toBe(path);
+  });
+
+  it("does not guess between ambiguous tool paths", async () => {
+    const resolved = await resolveOpenablePath(cwd, "backup.yaml", [
+      "/Users/me/one/backup.yaml",
+      "/Users/me/two/backup.yaml",
+    ]);
+    expect(resolved).toBe(`${cwd}/backup.yaml`);
+  });
+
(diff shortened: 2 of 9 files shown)


hardbeat920:
Thanks @yankawai the intent makes sense but I noticed two issues before merging:

1. filePaths is recreated whenever blocks changes, which recreates handleOpenFile. Since sameActivity compares 
onOpenFile by identity, every visible historical activity group will rerender during streaming. Could this use a ref or another stable-callback approach so the existing long-transcript optimization remains effective? 

2. transcriptFilePaths only collects preview.path. Codex multi-file edit events provide all affected files in event.paths, but those paths are not retained when the event becomes a transcript block. Consequently, shortened links for every file except the first one can still resolve incorrectly. Could we preserve all structured paths and add a multi-file test?

One smaller follow-up, right-click actions such as “Open in Default App” and “Reveal” bypass 
onOpenFile, so they still use the unresolved session-relative path.

The overall approach looks good once these cases are addressed.

yankawai:
All three are in, on top of today's main.

1. The callback no longer depends on the scan: the transcript keeps blocks and cwd in a ref, `handleOpenFile` has empty deps, and the paths are collected when a link is clicked, so streaming leaves `sameActivity` intact.
2. `tool.started` and `tool.updated` carry `paths`, and the row keeps them, so every file of a multi-file edit answers for its own link — a later progress patch no longer drops the list. Covered by a reducer test and a transcript test.
3. Reveal, Open in Default App and the copy actions go through the same resolution a click gets, instead of acting on the link text.

Merged main in rather than rebasing, so the review stays attached. Five new cases here; four of them fail with the fixes reverted.

PR #182, opened 2026-09-12
Keep the composer's height across tab switches

goujandev:
## What changed

The composer no longer writes a height it measured while its tab was hidden, and
re-measures when the tab becomes visible again.

## Why

Fill a prompt box to several lines, switch tabs, switch back, and it collapses
to one row with the draft stranded behind a scrollbar. Hidden tabs stay mounted
under `display: none`, so `scrollHeight` reads `0` and the resize wrote
`height: 0px`; nothing re-measured on the way back.

Opening the Inbox reproduces it. Settings does not, because `visible` ignores
`settingsOpen`, so `SessionPane` never re-renders and the bad measurement never
fires.

`GitChangesPanel` already guards the same pattern via `enabled`
(`src/chrome/GitChangesPanel.tsx:288`); this applies it to the composer.

## UI

Before: 

https://github.com/user-attachments/assets/1b8a7e99-4af7-41c9-982e-aa75ea9f47c1

After: 

https://github.com/user-attachments/assets/d556025e-cce1-4b8e-a39f-564627ca72eb

## Checklist

- [x] I ran `npm run check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved composer resizing as text is entered or content changes.
  * Composer height now grows to fit content while remaining within a maximum size.
  * Improved resizing when switching views or using the composer in panels where layout may load later.
  * Preserved the current composer height when its layout is temporarily unavailable.

* **Tests**
  * Added coverage for composer growth, maximum height limits, and unavailable layouts.

### DIFF
--- a/src/chrome/Composer.tsx
+++ b/src/chrome/Composer.tsx
@@ -35,6 +35,7 @@ import {
   pickAttachments,
   revokeAttachment,
 } from "../lib/attachments";
+import { resizeComposer } from "../lib/composerResize";
 import type { ContextUsage } from "../lib/contextUsage";
 import {
   loadProjectFiles,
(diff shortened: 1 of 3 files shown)


goujandev:
Note on the diff size: the height logic used to live in a local `resizeTextarea`
helper inside `Composer.tsx`. Moving it into `src/lib/composerResize.ts` so it
could be unit-tested meant renaming it to match the module, which accounts for 9
one-word call-site changes. None of them change behaviour. I could have exported
it as `resizeTextarea` to keep `Composer.tsx` to 3 hunks instead of 11, happy to
do that if you'd rather have the smaller diff.

hardbeat920:
@goujandev thank for this. merging :)

PR #197, opened 2026-09-12
Add agents panel with subagent tabs and live activity

RayenTellissy:
## What this adds

**An Agents panel.** There is a new robot icon in the title bar, next to the terminal button. Clicking it opens a panel on the left that lists every subagent the current sessions have spawned in their latest turn. Subagents are grouped by the session that started them. Running ones are listed first with the same shimmering loader the transcript uses; finished ones show a green check, failed ones a red cross. A dot on the icon shows when any subagent is still working. Clicking a session name jumps to that session.

**Each subagent line shows what it is doing.** Under the title you see the model it runs on (for example "Haiku 4.5" or "Opus 4.8") and its most recent action, such as the file it is reading or the search it is running.

**Subagents open in their own tab.** Clicking a subagent opens it as a tab next to the session, the same way plans do. The tab shows the subagent's title, model and status, then a live list of every tool it called and every message it wrote, updating as it works. When it finishes, the tab also shows its final result or error. Closing or archiving the session closes these tabs too, and they survive restarts.

**Subagent names stay put.** A subagent now keeps the short description it was spawned with, like "Find where tabs are reordered", instead of being renamed by later progress reports.

## How it works

- The Claude harness now attaches each subagent's model, tool calls and text output to the parent Agent block as a list of steps. Text still stays out of the main transcript.
- A new `subagents` module derives the panel contents from session blocks.
- A `subagent` tab kind was added to the workspace layout, with snapshot persistence and cleanup when the session goes away.
- Unknown model ids are prettified for display (`claude-opus-4-8` becomes "Opus 4.8").

Steps and model are only captured for the Claude harness for now. Subagents from other harnesses still appear in the panel but their tab shows no activity.

## Testing

- `tsc --noEmit` clean
- Full vitest suite passes (1407 tests), including new tests for step merging, title freezing, tab keys, and the harness capturing subagent model, tools and text.

## Summary by CodeRabbit

- **New Features**
  - Added an Agents panel showing grouped subagents, live running counts, models, statuses, and activity.
  - Open subagents in dedicated tabs to view progress, results, or errors.
  - Added clearer subagent tab names, icons, and tooltips.
  - Subagent activity now displays progress steps and model information as work proceeds.
  - Added tab actions for opening files externally, revealing or copying paths, and closing tabs.

- **Bug Fixes**
  - Preserved meaningful subagent titles during progress updates.
  - Removed subagent tabs when their associated session is removed.
  - Improved handling of completed, failed, and running states.


hardbeat920:
Thanks for the thoughtful work here @RayenTellissy 

We recently added a different subagent experience in 08ba69d, 283dab1, and c623308. 

It now handles activity trails, models, results, and multiple providers. 

This PR overlaps heavily with that implementation, so I am closing it as superseded. 
We may revisit the global Agents panel idea separately in the future so feel free to open a seperate pr for that

thank you again 🙏

PR #199, opened 2026-09-13
Gate Claude Sonnet 5 behind minimum CLI version

nulljosh:
## Summary
- Opus 5, Fable 5, Opus 4.8 and 4.7 are all excluded from the Claude model picker on CLI versions that predate them, but Sonnet 5 had no such gate in `modelsForClaudeVersion`.
- On older `claude` CLI installs, Sonnet 5 still showed in the picker and got passed via `--model`, which the CLI rejected as unrecognized ("There's an issue with the selected model...").
- Added `MINIMUM_CLAUDE_SONNET_5_VERSION` alongside the existing Opus 5 gate and filtered it the same way.

## Test plan
- [x] `npx vitest run src/lib/harness/registry.test.ts src/lib/harness/claudeProtocol.test.ts src/lib/harness/claudeLive.test.ts` — 55 passed
- [x] Reproduced locally: Sonnet 5 failed to start a session while Haiku 4.5 worked, matching the missing-gate hypothesis

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added support for Claude Sonnet 5 when using Claude version 2.1.197 or later.
  - Sonnet 5 now appears in the available model list when the installed Claude version meets the required minimum.

- **Bug Fixes**
  - Prevented Claude Sonnet 5 from appearing on unsupported Claude versions, including versions earlier than 2.1.197 or when the Claude version cannot be determined.

### DIFF
--- a/src/lib/harness/claudeCatalog.ts
+++ b/src/lib/harness/claudeCatalog.ts
@@ -24,6 +24,7 @@ import {
   MINIMUM_CLAUDE_OPUS_4_7_VERSION,
   MINIMUM_CLAUDE_OPUS_4_8_VERSION,
   MINIMUM_CLAUDE_OPUS_5_VERSION,
+  MINIMUM_CLAUDE_SONNET_5_VERSION,
   parseClaudeVersion,
   parseControlResponse,
   parseJsonLine,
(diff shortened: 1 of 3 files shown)


hardbeat920:
Thanks @nulljosh the approach makes sense, but I think the minimum version should be 2.1.197, not 2.1.219. Anthropic’s release notes say Sonnet 5 was introduced in 2.1.197; 2.1.219 introduced Opus 5.

Could we also test that 2.1.196 hides Sonnet 5 and 2.1.197 shows it?

nulljosh:
Good catch, thanks. Pushed a fix: gated `MINIMUM_CLAUDE_SONNET_5_VERSION` to `2.1.197` separately from Opus 5's `2.1.219`, and added the 2.1.196/2.1.197 boundary tests you asked for. All 56 tests pass.

hardbeat920:
@nulljosh thank you. Merging :)

PR #208, opened 2026-09-13
Choose an available provider for new sessions

notsapinho:
## What changed

New sessions now choose an installed provider when the saved/default provider is unavailable. The initial blank session is corrected after provider discovery, creation paths share the same availability-aware selection, and unavailable providers cannot be made the effective default in Settings.

## Why

A clean install could open on Cursor Composer 2.5 even when Cursor was missing and Codex or OpenCode was installed. Settings also presented unavailable providers as valid defaults.

## UI

Provider Settings now disables the default action for unavailable CLIs and labels the effective installed provider as the default.

## Checklist

- [x] I ran `npm run check:web`

## Summary by CodeRabbit

- **New Features**
  - New sessions automatically select an available provider when the preferred provider is unavailable.
  - Empty sessions using unavailable providers are updated at startup, while active, resumed, and transferred work remains unchanged.
  - Availability-aware defaults apply when creating sessions, inboxes, splits, panes, history entries, and managing projects.
  - Editor-only resets preserve the configured provider when available and otherwise select an available provider.
  - Provider settings show the effective default provider and disable “Use by default” for unavailable providers.


src/App.tsx
```diff
@@ -165,6 +165,7 @@ import {
   compactHarnessContext,
   forgetHarnessSession,
   generateHarnessTitle,
+  isHarnessAvailable,
```
hardbeat920: Current `main` already imports this symbol from `./lib/harness/availability`, so the rebased branch now declares it twice and fails `tsc` on all three CI platforms.

```suggestion

```

PR #220, opened 2026-09-13
Add Windows toast notifications via WinRT

ardevdevts:
## What changed

Implements the Windows backend for the existing notification system: session finished/needs-input banners (plus due reminders) now dispatch through the WinRT toast API via tauri-winrt-notification, with click-to-open-session and a Settings entry that deep-links to Windows notification settings. macOS/Linux behavior is unchanged.

## Why

The notification commands returned Unsupported on Windows, so turn-finished and reminder banners silently fell back to the in-app cue only. Windows 10/11 can show native toasts like the other platforms once an AppUserModelID path exists.

## UI

No layout change. The Settings Notifications blocked row now also offers its Open System Settings button on Windows (previously mac-only); the button opens ms-settings:notifications.

## Checklist

- [x] I ran the full check suite (vitest, tsc --noEmit, cargo fmt --check, cargo clippy -D warnings, cargo test) - all green

## Summary by CodeRabbit

* **New Features**
  * Added Windows notification permission detection, including per-app and system-level notification blocks.
  * Added an “Open System Settings” option when Windows notifications are disabled.
  * Reminder actions triggered from Windows notifications can open the related reminder.

* **Bug Fixes**
  * Clicking a notification now restores a minimized window before focusing it.
  * Windows notification delivery now stops cleanly when notifications are blocked in system settings.


hardbeat920:
@ardevdevts thanks for the work. Overall this is great however I found two issues before merge:

1. The backend always returns `Granted`, so the Windows “Open System Settings” button never appears when notifications are blocked. It also means a blocked toast may be treated as delivered, skipping the in-app cue.
2. Clicking a notification while MonoCode is minimized only calls `setFocus()`. On Windows, the window must be unminimized first, otherwise it remains minimized.

Could you address these cases? 🙏

hardbeat920:
@ardevdevts Thanks for the updates. I checked the latest changes and both issues are resolved. Looks good to me.

PR #226, opened 2026-09-14
Autofocus search when opening the sidebar project picker

### LINKED ISSUE #225: Enhancement: autofocus search when opening the sidebar project picker
## Summary

When the project picker in the sidebar opens, focus doesn't go to the "Search projects..." input. You have to click the field before you can type to filter projects.

## Proposed behavior

When the picker opens, the search input should be focused so you can type straight away, the same as other search-first pickers.

## Steps to reproduce current behavior

1. Open MonoCode (0.1.45, Linux).
2. In the sidebar, click the project switcher to open the project picker.
3. Start typing: the keystrokes don't reach the search input because it isn't focused.

actuallyakshat:
Closes #225

## What changed

The search input in the sidebar project picker now gets focus when the picker opens. It's focused through a ref when `open` becomes true, and again on the next animation frame. A regression test covers it.

## Why

When you opened the project picker, the search field wasn't focused, so you had to click it before typing. The `autoFocus` attribute on the input wasn't reliably taking effect when the picker opened.

## UI

No layout or styling change. The only difference is that the search field is focused as soon as the picker opens, so you can start typing right away.

monocode auto search

## Checklist

- [x] I ran `npm run check:web` (vitest + `tsc --noEmit`, both green). No Rust files are touched; I couldn't run `check:rust` locally because `rustfmt` isn't installed.

## Bug Fixes

- Improved the project picker so its search field receives focus when the picker opens.
- Added a reliable focus retry to restore focus if it is temporarily lost during loading or rendering.
- Added coverage to verify that focus is restored after the deferred retry.


hardbeat920:
@actuallyakshat thank you thats a great detail

PR #227, opened 2026-09-14
feat: per-session Claude account profiles

### LINKED ISSUE #198: Support multiple Claude Code accounts/profiles (only the default ~/.claude login is reachable)
## Problem

MonoCode's Claude harness resolves and drives exactly one `claude` binary — the one it finds at a fixed set of paths / via a login-shell PATH lookup (`src-tauri/src/harness.rs::resolve_claude`). There's no way to point MonoCode at a *different* Claude Code account/profile than whatever is currently the default at `~/.claude`.

This matters for anyone juggling multiple Claude Code logins today (e.g. a work account and a personal account), which people commonly do via shell aliases like:

```sh
alias claude-work='CLAUDE_CONFIG_DIR=~/.claude-work claude'
alias claude-personal='CLAUDE_CONFIG_DIR=~/.claude-personal claude'
```

None of this reaches MonoCode:
- Aliases only exist inside an interactive shell session — they're invisible to `resolve_claude`'s file-based lookup.
- MonoCode *does* shell out to a login+interactive shell (`zsh -lic printenv`) to read a few env vars, but that's a hardcoded 6-key allowlist (`PATH`, `AI_GATEWAY_API_KEY`, `FX_AI_GATEWAY_API_KEY`, `VERCEL_OIDC_TOKEN`, `XAI_API_KEY`, `GROK_CODE_XAI_API_KEY`) unrelated to Claude — so even setting `CLAUDE_CONFIG_DIR` in `.zshrc` never reaches the spawned `claude` process.
- The harness registry (`registerBuiltinHarnesses`) only has one slot per provider id — there's no concept of multiple named instances of the same provider.

Net effect: whichever account is the unaliased default at `~/.claude` is the only one usable from inside MonoCode. The other accounts are only reachable from a regular terminal, outside the app.

## Suggested direction

Some form of a per-provider account/profile picker for Claude (and potentially other providers with the same multi-login pattern), e.g.:
- Let a project (or a session) pin a specific `CLAUDE_CONFIG_DIR` (or equivalent) to use when spawning `claude` for it, instead of always relying on the environment default.
- Surface configured profiles in the same place model/provider selection already lives, rather than requiring the user to change global default credentials and restart the app to switch.

I don't have a concrete design opinion yet — flagging this as a real gap I hit, per CONTRIBUTING.md's guidance to open an issue before attempting anything that changes product shape. Happy to discuss the approach or pick it up if it's welcome.

Nishanth-sebastin:
Closes #198.

Adds named per-session/per-project Claude account profiles, each pointing at a different `CLAUDE_CONFIG_DIR` — the gap described in #198 (only the default `~/.claude` login was reachable from inside the app).

- Named profiles, picked per session/project, each resolving to its own `CLAUDE_CONFIG_DIR`
- Usage footer now reads (and lets you switch) the active session's own account, per the direction discussed in the issue
- Fixes a conversation-resume bug that surfaced across profile switches (stale thread IDs were being resumed against the wrong account)

Verified: `tsc --noEmit` clean, full vitest suite (506 passing), `cargo check`/`fmt --check`/`clippy -D warnings`/`test` (250 passing) all clean.

Happy to adjust the UX/naming if you had something more specific in mind for the footer picker.

## Summary by CodeRabbit

- **New Features**
  - Added Claude account profile selection, including Personal and additional configured profiles.
  - Profiles can be available based on the current project folder.
  - Claude sessions preserve the selected profile when resuming.
  - Added project terminal controls and provider sign-in/reconnection flows.
  - Added Codex usage-credit reset support.

- **Bug Fixes**
  - Claude usage limits now reflect the active session’s account.
  - Switching accounts refreshes usage data and starts a separate session when needed.
  - Prevented sessions from resuming when their folder or account profile no longer matches.
  - Expired Claude credentials now request sign-in instead of refreshing automatically.


hardbeat920:
Thanks for putting this together @Nishanth-sebastin and for helping shape the account-switching direction, especially the account-aware resume and usage behavior. 

#280 has now merged and covers the core feature for both Claude and Codex, so I’m closing this PR as superseded.

If you’d like to keep improving this area, focused follow-up PRs against the current main would be very welcome, for example, any remaining UX gaps, importing existing profiles/config directories, migrations, or edge cases you notice while using #280.

PR #233, opened 2026-09-14
Fix Windows shell command presentation

notsapinho:
## Summary
- unwrap PowerShell and Pwsh command launchers for readable tool rows
- unwrap cmd /c launchers while preserving POSIX shell behavior
- retain incomplete commands and quoted flag-like arguments unchanged

## Verification
- npm run check:web (193 files, 2219 tests)
- TypeScript typecheck included in check:web

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shell command detection and unwrapping for POSIX shells, PowerShell, and Command Prompt.
  * Added support for common executable aliases and quoted launcher options.
  * Correctly preserves trailing arguments and command remainders after wrapped commands.
  * Prevents PowerShell script arguments from being mistaken for wrapper options after `-File` or `-f`.
  * Preserves ordinary, incomplete, and unclosed-quote commands without altering them.


hardbeat920:
Thanks @notsapinho 

Could you tighten the quote handling before we merge?
trimMatchingOuterQuotes() currently removes the first and last quote without confirming they enclose the whole command. 

For example:
`pwsh -Command 'Get-Date' '-Format' 'yyyy-MM-dd'`

becomes:
`Get-Date' '-Format' 'yyyy-MM-dd`

Also, quoted options such as `pwsh "-Command" "Get-Content package.json"` are not unwrapped because quoted tokens are skipped, even though the shell passes "-Command" as the normal -Command argument.

Could you preserve separately quoted arguments, recognize quoted wrapper options while still stopping at -File/-f, and add tests for both cases? 

Everything else looks good and CI is green.

hardbeat920:
Thank you this looks great 🙏

PR #237, opened 2026-09-15
Fold a settled turn's interjections, status rows, and delegated runs into the work trail

elijah7x:
## Summary

Settled turns currently leak their process all over the transcript: every OMP interjection stays a labeled divider card, every status row splits the work groups, and each delegated run keeps a pinned row forever. A busy turn can leave dozens of standalone chrome rows between the question and the answer.

This change makes the fold time-aware instead of adding more exceptions:

- **Live turns are untouched.** Interjections still land on their own labeled rows, status rows still mark the stream, and delegated runs keep their pinned rows — the reader needs somewhere to watch.
- **Settled turns fold all of it into the work trail.** Interjections and delegated runs join the activity groups; ordinary status rows are work steps; the transcript reads as prompt, one fold line, and the answer.
- **Inside the trail everything stays reachable:** advisor/IRC notes become compact one-line rows that open to the full note; each delegated run is a compact row that opens onto its own trail, unchanged.
- **A run that died keeps its own row** — a rejected delegated run parks under the fold line already open on the provider's reason, exactly as it did while live.
- **Folded prose is muted** (`zen-fold-prose`): dimmed ink, headings demoted to bold lines, tighter spacing — so mid-turn notes can no longer pose as the final answer.

Summaries gain a notes clause (`Ran 3 commands · 2 notes`), and a group holding nothing but absorbed notes is just that clause. Status rows count nowhere.

## Test plan

- [x] Unit matrix for the settled trail: status rows in-group (live/default/settled), interjection live-standalone vs settled-folded, delegated run live-pinned vs settled-folded, failed run kept outside the fold, fold spanning the whole trail, trailing notes after the answer, fold reaching across a live-fold-stopping interjection
- [x] Component tests: live divider vs settled compact note row (opens to full text), failed run auto-open in both modes, `zen-fold-prose` marker lands on folded prose only
- [x] Rendered three real persisted sessions (up to 40 interjections / 5 delegated runs / 106-block folds): settled DOM shows zero standalone dividers or subagent panels; live DOM keeps the last turn's 11 interjections as their own rows
- [x] `npm run check:web` — 2221 tests green; `npx tsc --noEmit` clean

## Summary by CodeRabbit

- **New Features**
  - Improved folded work-trail summaries with clearer note, status, command, and delegated-run information.
  - Added compact, labeled expandable rows for interjections and system status updates.
  - Failed delegated tasks now remain visibly expanded so failure details are easier to review.
  - Settled work trails group related activity more clearly while preserving important notes and rejected runs.
  - Refined Markdown styling within folded work areas for improved readability.

- **Bug Fixes**
  - Improved consistency between live and completed transcript displays, including interjections and delegated activity.
  - System errors and interruptions now remain clearly distinct from regular activity summaries.


hardbeat920:
@elijah7x can you please share some screenshots as it is hard for me to replicate. that would be very helpful 🙏

elijah7x:
Sure — here's a synthetic session reproducing the pattern that's hard to trigger locally: OMP advisor interjections, `Advisor reviewed` status rows and delegated subagent runs all in one turn. Same stored blocks rendered before and after this PR.

**Before** — the settled turn stays fragmented: every advisor divider, status row and delegated run breaks the work trail apart:

pr237-before

**After (collapsed)** — the same turn settles into a single work-trail line; the final answer stays outside the fold:

pr237-settled-collapsed

**After (expanded)** — expanding the trail keeps every step reachable as compact rows: tool calls, advisor notes, `irc:incoming` messages and subagent runs (each still individually expandable). Prose inside the fold is muted so it reads as process, not the answer:

pr237-settled-expanded

**Live is unchanged** — while the turn is running, interjection dividers and delegated runs stay pinned exactly as before; folding only applies once the turn settles:

pr237-live

One more detail: a failed delegated run still renders as its own auto-expanded row outside the fold, so failures are never buried.

hardbeat920:
that looks great. I review asap 🙏

hardbeat920:
Looks like a good improvement overall, but there’s one issue to address before merging.

All non-interjection system blocks are now treated as foldable status messages. Provider errors use the same block type, so an error following a tool call can be folded under “Ran a command” and effectively hidden.
Please keep errors and interruption notices visible, and add a test for user → completed tool → system error
Also, status-only groups currently show as “Thought” a status-specific label would be clearer.

Other than that it looks great :)

elijah7x:
Both points addressed in 2de9767 — thanks for catching the error case.

**Notices stay visible.** Provider errors, the quit-interrupt note and the not-connected message now carry a `notice` tag on the block (produced at the source, survives persistence), and `isActivityBlock` leaves them as standalone rows — live and settled alike, and they bound the fold so they can't be swept under it either. Interrupts persisted before the tag existed are still recognised by their literal text, so old sessions stay fail-visible too. Added the user → completed tool → system error test for both live and settled.

**Status-only groups** are now labelled "Status update" instead of falling through to "Thought".

`check:web` green: 188 tests across the transcript/harness/store suites + tsc clean.

hardbeat920:
Thanks, this addresses the main issue I found. The added persistence and regression coverage also looks good.

There’s still a small merge conflict with current main, plus the early return in the delegated-run test should ideally fail instead. 

Otherwise this looks good to me.

hardbeat920:
Quick scope check... this folding logic is in the shared transcript and isn’t gated to OMP, so status rows, delegated runs, and folded prose will change for every harness. Only the advisor/interjection data appears OMP-specific. 

Is that broader UI change intentional?

elijah7x:
Done — merged current main (trivial import conflict in AgentTranscript), the delegated-run test now fails instead of returning early, and a status-only group reports `workKind: "note"` so it takes the note presentation. 2139 tests + tsc green.

On the scope question — yes, intentional. The folding is driven by the turn lifecycle (live vs settled), not by anything OMP-specific: status rows and delegated runs exist for every harness, and a long Codex/Claude session produces the same trail noise. OMP is just what made it obvious. Interjection *content* stays generic chrome (`customType` is already harness-agnostic — `advisor`, `irc:incoming`, whatever a provider emits); the PR only changes when that chrome folds, never what it renders.

If you'd rather keep the new behaviour OMP-gated I can scope it, but non-OMP sessions would keep the fragmented settled transcript — the cleanup is the same either way.

hardbeat920:
Looks good to me. Thanks for addressing the previous concerns. The remaining accessibility label issue is minor and i can handle it in a follow-up 

Thanks again 🙏

PR #239, opened 2026-09-15
Space agent-markdown blocks through Streamdown's dir wrappers

### LINKED ISSUE #218: Replies render as one dense block: blank lines between paragraphs are lost
### MonoCode version

0.1.44

### Operating system

macOS

### OS version

macOS 26.6.2

### Provider CLI

Codex

### What happened

**What I expect**

I keep a global instruction file, and its formatting rules are explicit:

- Separate every paragraph, heading, and list block with a blank line.
- Never separate two blocks with a single newline. A single newline is treated as part of the same paragraph, so the reply renders as one dense block with no spacing.
- Prefer one idea per paragraph over one sentence per line.

**What happens**

Assistant replies are shown as one continuous block. In the attached screenshot, three separate sections run together with only ordinary line spacing:

- On performance: nothing runs per move. The panel reports only on hover enter and exit...
- Tests: 128 pass, no warnings. The app is relaunched on this build...
- I cannot move the mouse to verify hover myself, synthetic events are off limits here...

Screenshot of three sections running together with no paragraph spacing between them

A reader has to parse where one idea ends and the next begins. It is the specific failure the instruction above exists to prevent.

**Evidence that the text itself does contain the breaks**

I read the stored messages for the sessions in question. For four long replies on 2026-09-13:

| Reply length | Blank-line breaks (`\n\n`) | Single-newline breaks |
| --- | --- | --- |
| 3022 | 17 | 25 |
| 5433 | 20 | 40 |
| 11689 | 30 | 91 |
| 9677 | 24 | 68 |

The paragraph breaks are present in the data. They are not visible as paragraph separation in the transcript.

**What I would like you to check**

Whether the assistant text path renders Markdown paragraphs with real paragraph spacing, or displays the text as raw pre-wrapped content where `\n\n` collapses to one blank line. If it is a `<pre className="whitespace-pre-wrap">` surface, `\n\n` will only ever read as a single blank line no matter what the model emits.

**It is specific to MonoCode**

With the same model, the same provider, and the same local proxy, Hermes and the Codex app both render the same instruction correctly. Only MonoCode collapses it.

### Steps to reproduce

1. Set the instruction above in your global Codex instructions file.
2. Run any multi-section turn that asks for several points of discussion.
3. Read the reply in the transcript.

For comparison, run the same turn with the same model and provider through Hermes or the Codex app. Both keep the paragraph separation, MonoCode collapses it.

bluzername:
Fixes #218

Updated after @hardbeat920 could not see a difference. He was right: the first patch on this branch changed nothing. Details and screenshots are in the comment below.

## What happens

A reply with several "\n\n"-separated sections shows up as one dense block in the transcript, no gap between the paragraphs, even though the breaks are really there in the stored text.

## Cause

`AgentMarkdown` passes `dir="auto"` to Streamdown. With that, Streamdown puts every block in its own wrapper:

```html
<div class="space-y-4 ... agent-markdown ...">
  <div dir="ltr" style="display:contents"><p>First paragraph.</p></div>
  <div dir="ltr" style="display:contents"></div>
  <div dir="ltr" style="display:contents"><p>Second paragraph.</p></div>
</div>
```

A `display: contents` box drops its own margins. Streamdown's `space-y-4` on the root targets exactly those wrappers, so it paints no gap at all. Headings, blockquotes and rules still look fine because their margin sits on the element itself (`mt-6 mb-2`, `my-4`, `my-6`). Paragraphs and lists carry no margin, so they sit flush.

The same wrapper is why my first patch did nothing: every `<p>` is the only child of its wrapper, so `.agent-markdown p:last-child` matched all of them and zeroed the `margin-bottom` it had just added.

## Fix

Put the gap on the block inside the wrapper, top side only, so a reply never ends with a trailing gap:

```css
.agent-markdown > div + div > p,
.agent-markdown > div + div > [data-streamdown="unordered-list"],
.agent-markdown > div + div > [data-streamdown="ordered-list"] {
  margin-top: 1rem;
}
```

Lists are in there because they have the same problem, no margin of their own. Headings, blockquotes and rules are left alone, they already space themselves.

Measured in Chromium at 860px wide, dark theme, rendering the real `AgentMarkdown` with the real `src/index.css`:

| boundary | main | this branch |
| --- | --- | --- |
| paragraph to paragraph | 0px | 16px |
| paragraph to list | 0px | 16px |
| above the first block | 0px | 0px |
| below the last block | 0px | 0px |
| heading to paragraph | 8px | 16px |

The heading row moves because the paragraph's new 16px collapses with the heading's own 8px bottom margin. That is the 1rem Streamdown meant to have there anyway.

## Test

`src/surfaces/agentMarkdownSpacing.test.ts` now renders real Streamdown output (`react-dom/server`, with `// @vitest-environment happy-dom` in the file so the repo-wide node environment stays as it is) and checks the shipped selectors against that DOM. So a rule that matches nothing, or matches every paragraph like my first one did, fails. It is red on main, red on my first patch, green now.

Ran `vitest run` (2223 tests green) and `tsc --noEmit` (clean). Did not run the cargo half of `npm run check`, nothing here touches Rust.

## Summary by CodeRabbit

- **Style**
  - Improved readability of agent-rendered Markdown by adding spacing between consecutive paragraphs, ordered lists, and unordered lists.
  - Prevented extra spacing after the final content block for a cleaner layout.

- **Tests**
  - Added coverage for spacing between consecutive content blocks.
  - Verified that the first block has no preceding spacing and the final block has no trailing spacing.
  - Added checks for different margin declarations and invalid selectors.


hardbeat920:
@bluzername can you share some screenshots to see the difference. I honestly couldn't find any but maybe i am missing sth 🙏

bluzername:
thanks for checking. You are right, and I owe you an apology: my first patch did nothing at all. I rendered both and the two screenshots came out byte identical (same file size, `cmp` says no difference). So there was nothing to see.

Here is why, and what I found.

`AgentMarkdown` passes `dir="auto"` to Streamdown. With that set, Streamdown puts every block in its own wrapper:

```html
<div class="space-y-4 ... agent-markdown ...">
  <div dir="ltr" style="display:contents"><p>First paragraph.</p></div>
  <div dir="ltr" style="display:contents"></div>
  <div dir="ltr" style="display:contents"><p>Second paragraph.</p></div>
</div>
```

Two things follow from that wrapper:

1. A `display: contents` box drops its own margins. Streamdown's `space-y-4` on the root targets exactly those wrappers, so it paints no gap. That is the real cause of #218. Headings, blockquotes and rules still space fine because their margin sits on the element itself.
2. Every `<p>` is the only child of its wrapper, so `p:last-child` matches all of them. My `.agent-markdown p:last-child { margin-bottom: 0 }` cancelled the `margin-bottom: 1rem` I had just added on the line above. A perfect no-op. That is what you were looking at.

New commit on the branch puts the gap on the block inside the wrapper instead, top side only, so no trailing gap:

```css
.agent-markdown > div + div > p,
.agent-markdown > div + div > [data-streamdown="unordered-list"],
.agent-markdown > div + div > [data-streamdown="ordered-list"] {
  margin-top: 1rem;
}
```

Before (main, and also my first patch, identical):

before

After (this branch now):

after

What is in the screenshots: the real `AgentMarkdown` component with the real `src/index.css` and the real Tailwind build, served by `vite`, driven with Playwright in Chromium. 860px viewport, default dark theme, 760px content width, the three-paragraph sample from #218. Only `src/index.css` differs between the two shots.

Measured with `getBoundingClientRect` on the `<p>` elements:

| | main | first patch | now |
| --- | --- | --- | --- |
| gap between paragraph 1 and 2 | 0px | 0px | 16px |
| gap between paragraph 2 and 3 | 0px | 0px | 16px |
| computed margin on each `<p>` | 0px / 0px | 0px / 0px | 16px top |
| block height | 144px | 144px | 176px |

The test is rewritten too. It now renders real Streamdown output with `react-dom/server` and checks the shipped selectors against that DOM, so a rule that matches nothing, or matches every paragraph like my first one, fails. It is red on main, red on my first patch, green now. It uses `// @vitest-environment happy-dom` in the file, so the repo-wide node environment in `vitest.config.ts` stays as it is. `vitest run` 2223 green, `tsc --noEmit` clean, prettier clean.

Sorry for the round trip. Thank you for pushing back instead of merging it.

hardbeat920:
@bluzername thank you this looks good. Merged :)

PR #240, opened 2026-09-15
Composer effort control: continuous quantum-light slider

elijah7x:
## What changed

Replaces the dedicated composer effort control's radio list with a horizontal slider: a continuous rail with deterministic quantum specks, discrete tier ticks, and a 27px thumb that tracks the pointer 1:1 and snaps to the nearest tier on release. The composer chip shows a miniature version of the same rail.

## Why

Effort is a discrete, model-specific choice (3–7 tiers depending on the harness, e.g. Claude's low → ultrathink), but selecting it through a menu hides its ordered nature. The slider makes the ordinal structure visible — more effort reads as more light — while keeping each option a real tier: drag motion is continuous, commits are always snapped to an actual model option.

Design decisions worth knowing:

- **Grayscale quantum-light language**: the fill is a fixed full-width luminance gradient revealed by clip-path, so the leading edge is always the brightest point; 30 deterministic specks inside the fill brighten with effort level and respond locally to the thumb (a velocity-scaled trailing glow while dragging, a faint lift while held).
- **Top tier gets a restrained accent**: only the final 28px of the rail takes a 12% accent tint plus a soft thumb aura, honoring `--user-accent-color`.
- **Drag feel**: pointer pick-up preserves the grab offset inside the thumb (no jump when grabbing an edge); release settles over a distance-aware 110–190ms ease with no positional overshoot; thumb deform stays subpixel.
- **Ticks as landmarks**: 1×3px marks masked away near the thumb; `auto` renders as a hollow dot.
- **Fallbacks**: `prefers-reduced-motion` keeps direct tracking but disables twinkle, highlight, and deformation; the effort list inside the model picker's settings menu is unchanged.
- Accessibility: `role="slider"` with value text per tier, Arrow/Home/End keys, wheel stepping, Escape/Enter to close, reset-to-default button, and a focus keyline on the thumb (WKWebView's platform focus ring is suppressed deliberately).

## UI

Before: menu-style radio list in the popover. After: luminous rail + thumb (dark/light, mid and top tier verified). Happy to attach screenshots or a screen recording — the effect is motion-based and reads much better live.

## Checklist

- [x] I ran `npm run check`

## Summary by CodeRabbit

* **New Features**
  * Added an interactive effort meter for selecting ordered effort tiers.
  * Added progress indicators, tier highlights, compact previews, accent colors, focus states, and reduced-motion support.
  * Supports keyboard navigation, mouse-wheel stepping, pointer dragging, snapping, reset, and Enter/Space to close.
  * Unsupported restored effort values now fall back to a valid available tier.
* **Accessibility**
  * Added slider semantics and accessible effort value labeling.
  * Effort selection opens in a dedicated dialog.


elijah7x:
A candid note on scope: this PR is a fairly opinionated, preference-driven redesign of the effort control — I built it because I like this style of strength visualization, not because it was requested. If the interaction language or the visual treatment cuts against the design contract you have in mind for the composer, please feel free to decline; no offense taken. Also happy to shrink it to a smaller piece (e.g. just the ordering/discrete-tier semantics, or the plain slider without the particle treatment) if any part is useful on its own.

CodeRabbit's three comments are addressed in e90dc96.

PR #253, opened 2026-09-16
Add reliable Inbox mark-all-as-read actions

### LINKED ISSUE #252: Add reliable mark-all-as-read actions to Inbox navigation
﻿## Problem

The Inbox toolbar marks items read for the selected provider, but the Inbox navigation menu has no action to clear unread items across providers. Users must open Inbox and visit each provider. A failed local-storage write is also silently ignored, leaving items unread without explaining why.

## Proposed solution

Add "Mark all as read" to the Inbox navigation context menu, including when the project rail is collapsed. Use already loaded Inbox activity so the action remains available when a provider is slow or offline, and include muted projects without changing notification preferences.

The action should preserve the latest known update when the same repository is opened from another local checkout. A failed write should display an error and allow retry in both the menu and Inbox toolbar. Later updates must become unread again.

## Acceptance criteria

- All known unread items in the current Inbox project scope can be marked read across GitHub, GitLab, and Linear.
- The action is disabled when there are no known unread items, and indicators react to successful reads.
- Changing local checkout paths does not exclude an unchanged item or overwrite a newer timestamp with an older snapshot.
- Failed persistence leaves unread state intact, reports the failure, and allows retry.
- Mouse and keyboard context-menu access work with the project rail expanded or collapsed.

## Additional context

The checkout-path edge case was reproduced while reviewing the proposed menu implementation; it is not a claim about a released menu feature. Regression tests cover this case and persistence failure/retry in both entry points.

ognjeeen:
﻿## What changed

Add "Mark all as read" to the Inbox context menu in both expanded and collapsed project navigation. The action uses loaded activity across providers, includes muted projects, and updates unread indicators without waiting for another provider request.

Keep all known local checkout associations for an item while retaining its newest timestamp. Report failed read-status writes in both the context menu and Inbox toolbar, preserve unread state, and allow retry.

Closes #252.

## Why

Users need a single navigation action to clear known Inbox unread items. An unchanged remote item must remain actionable after switching local checkout paths, and storage failures must not look like successful reads.

## UI

The Inbox context menu gains a "Mark all as read" entry. Failed writes show an inline error; the menu remains open for retry. Mouse, ContextMenu, and Shift+F10 behavior are covered by component tests. No manual desktop screenshots were captured.
image

## Validation

- Added three regression tests, each observed failing before its fix: checkout reassociation with an older snapshot, menu persistence failure/retry, and Inbox toolbar persistence failure/retry.
- `npm run check:web`: 216 test files, 2,374 tests passed; TypeScript passed.
- `npm run check:rust`: formatting, Clippy, and 248 Rust tests passed.
- `git diff --cached --check` passed before commit.
- Checks ran locally on Windows for commit 375d70e. The two check scripts were run separately and cover the commands in `npm run check`.

## Checklist

- [x] I ran the equivalent of `npm run check`, using `check:web` and `check:rust` separately

## New Features
- Added an Inbox context menu to the sidebar, available via right-click and keyboard shortcuts.
- Added actions for opening notification settings and marking all Inbox items as read.
- Added project-aware tracking for Inbox activity and unread status.

## Bug Fixes
- Improved synchronization of unread indicators when Inbox items change or are updated externally.
- Added visible error messages and retry support when saving read status fails.
- Preserved focus on the Inbox button after closing the context menu.


hardbeat920:
@ognjeeen looks good. Merging

PR #255, opened 2026-09-16
Rotate Codex accounts when a turn hits the usage limit

elijah7x:
## What changed

The Codex harness can now fail over to another ChatGPT account when a turn hits the usage limit instead of losing the session. A credential pool lives on the Rust side (atomic `0600` writes, redacted listing, serialized mutations), the session pins to an account, host-managed OAuth refresh answers `account/chatgptAuthTokens/refresh`, and failover is bounded: hot `account/login/start` in the same app-server when the candidate's identity verifies, respawn + `thread/resume` + turn replay when it does not. Shared workspace-level limits (`workspace_*`) are detected and reported rather than rotated.

## Why

Lately I keep hitting Codex quota walls - Astra/OMP sessions burn through an account fast, so I have to run several accounts. With a single account wired in, MonoCode's Codex session just dies at the wall and the turn is gone. This keeps the turn alive by rotating to the next account, the way OMP does.

Notes on scope: the pool is populated through the new `codex_account_upsert` command, and the existing `~/.codex/auth.json` login always remains a failover candidate. A settings UI for adding/removing accounts is deliberately left out to keep this reviewable; happy to send it as a follow-up. Account switching never calls `account/logout` (it deletes the on-disk auth state), and a rejected candidate is disabled and skipped rather than killing the turn.

## UI

No UI changes. A status line ("Codex account exhausted — switching to ...") appears in the transcript during failover.

## Checklist

- [x] I ran `npm run check`

## Summary by CodeRabbit

* **New Features**
  * Added Codex account management, including secure storage, credential refresh, account removal, and state tracking.
  * Added support for external Codex authentication and safe credential synchronization.
  * Added automatic account failover when quota limits are reached, including session recovery where possible.
  * Added rate-limit visibility and improved handling of account-related errors.

* **Bug Fixes**
  * Prevented duplicate session errors and unsafe request handling during failed turn startup.
  * Preserved authentication data and unknown fields during credential updates.


elijah7x:
Withdrawing for now: I want to validate the failover path against real Codex accounts first (the rotation chain is only covered by mocked tests so far), and the pool currently has no way to be populated from the UI. Will reopen or send a replacement once that is verified.

PR #256, opened 2026-09-16
Detect Codex auth refresh failures and show recovery steps

D3nnis72:
## What changed

Detects Codex auth refresh failures across all three event shapes (error event, failed turn, completed agent message) and renders actionable re-login guidance instead of the raw "Your access token could not be refreshed" text. Duplicate reports of the same failure within one turn collapse into a single row.

## Why

Codex refresh tokens are effectively single-use across parallel app-server processes sharing ~/.codex/auth.json. MonoCode spawns several of them (session, usage probe, model catalog), plus the Codex app and editor extensions do the same. One process wins the refresh, the others fail permanently until stale processes are restarted and the user signs in again (upstream openai/codex #9634). Before this change, the raw error appeared as plain assistant text, twice per turn, with no hint that re-login alone is not enough while stale processes keep running.

## UI

No layout or chrome changes. The failure previously rendered as two identical assistant text rows ("Your access token could not be refreshed. Please log out and sign in again."). It now renders as a single system row containing the original message plus recovery steps: quit other Codex clients, kill leftover `codex app-server` processes, run `codex logout` / `codex login` / `codex login status`, verify with one client first. No screenshots attached; verified via the new unit tests and a live session.

## Checklist

- [x] I ran `npm run check` (web: vitest + tsc green; full check with cargo not run locally, no Rust toolchain installed, no Rust code touched)

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate system error messages when identical session errors recur.
  - Streaming sessions now settle correctly when authentication errors occur.
  - Improved handling of Codex authentication-refresh failures during completed or failed turns.
  - Authentication failures now provide guidance to retry login or restart the app server.
  - Preserved normal agent messages that mention authentication issues without indicating an actual failure.
  - Preserved existing retry behavior for unrelated temporary errors.


hardbeat920:
Thanks for this @D3nnis72 

Just a couple of things before merge:
1. Only detect auth failures from structured errors or exact known messages. Normal agent text containing the phrase must remain a normal response.
2. Test the full streamed sequence and ensure only one error row remains.
3. Only deduplicate previous blocks when `notice === "error"`.
4. Make the recovery text less certain and account for Windows, where `pkill` is unavailable.

Thanks again 🙏

D3nnis72:
Thanks for the review! All four addressed:

1. Plain agent text now only rewrites on exact match against the known Codex messages (isKnownAuthRefreshMessage); anything merely mentioning the phrase stays a normal response. Broad matching remains only for structured error events and failed turns.
2. Added a streamed-sequence test (delta + completed + turn error) asserting exactly one error row.
3. Dedup now only collapses against a previous block with notice === "error", with a test proving status rows are untouched.
4. Recovery text hedged ("one possible cause") and covers Windows via Task Manager instead of pkill.

PR #257, opened 2026-09-16
Linux port + Mavis / MiniMax Code provider + glass UI (CodeRabbit fixes)

sahilcodexx:
Supersedes #140 — same scope plus the CodeRabbit review fixes.

## What's in this PR

Bundles the Linux port, mcode (Mavis / MiniMax Code) provider, glass UI,
and the dedicated AppImage CI job, plus all of the CodeRabbit review
fixes from #140.

## CodeRabbit fixes (since #140)

**🔴 Critical — `stringField` ignored key names.** The helper returned
the first non-empty string of the first record instead of the requested
key. `request.kind`, `request.title`, `request.callId`, and the ACP
session id all came back wrong — this could auto-allow write-class
tools in plan mode because the kind check was on a wrong field.
Split positional args into records + keys.

**🟡 Minor — plan events.** ACP plan updates carry an
`entries:[{content,priority,status}]` array. Switched to a proper
`tasks.updated` event (matching `fxProtocol.ts:planEvent`) so the UI
renders the plan as a todo list. Plain-text plan still works as a
fallback.

**🟠 Major — spawnChild placement.** Moved inside the existing try in
`ensureLive` so spawn failures reach the catch cleanup path that
calls `unwatchChild` / `stopMcodeSession`.

**🟡 Minor — handlePermission respond catch.** Added
`.catch(() => undefined)` to the `live.acp.respond` call so late
responses after teardown don't become unhandled rejections.

**🟡 Minor — install URL.** The fabricated `https://minimax.com/install.sh`
404s. Use the real URL `https://filecdn.minimax.chat/public/install.sh`
in both `availability.ts` and `harness.rs`.

**🟡 Minor — mcode_help_mentions_acp.** Only read stdout; merged
stdout + stderr to match `fx_help_mentions_acp` (some harnesses
print help only on stderr).

**🟡 Minor — Dark-mode lightness slider.** Last `glassDisabled`
reference; removed it so the slider works in both themes.

**🟡 Minor — docstring coverage.** Added JSDoc / \`///\` to every
new exported function in `mcodeProtocol.ts`, `mcode.ts`,
`mcodeAdapter.ts`, and the Rust `harness.rs` functions. Coverage
goes from 23.64% to comfortably above 80%.

**🟠 CI failure — npm ci lockfile out of sync.** Lockfile had older
deps; `pnpm install` regenerated it so the CI `npm ci` step
accepts the install.

## Linux port + mcode provider (initial PR contents)

- **pnpm setup.** `.npmrc` with shamefully-hoist + auto-install-peers
  + lockfile. Adds the CodeMirror core packages and mermaid which
  pnpm's strict layout otherwise hides.
- **Cross-platform Tauri config.** macOS-only `titleBarStyle` /
  `hiddenTitle` / `transparent` moved out of `tauri.conf.json`
  into `tauri.macos.conf.json`. `tauri.linux.conf.json` switches
  to transparent + decorations off + `backgroundColor: #00000000`
  so the desktop shows through. Also dropped the `resources: null`
  override that was clobbering the base config.
- **mcode harness.** `harness_resolve_mcode` Tauri command +
  `resolve_mcode` / `is_mcode_agent` / `file_mentions_mcode_agent`
  / `mcode_help_mentions_acp` Rust resolvers. Spawns `mcode acp`
  over stdio, drives it through the existing ACP client. Live
  adapter: initialize → session/new|resume|load → session/prompt,
  permission auto-approve, session/cancel. `mcode:minimax-m3`
  (native id `minimax/MiniMax-M3`) shows up in the picker
  alongside claude / codex / cursor / grok / opencode / pi / omp /
  fx. `HARNESS_ORDER` updated so the master index used by
  `findModel()` (and therefore the Favorites tab) actually
  surfaces it.
- **Glass UI cross-platform.** `HAS_NATIVE_GLASS = true` everywhere;
  the glass rules drop `:not(.theme-light)` so both themes get
  `backdrop-filter` on the sidebar and body pane. The blur-radius
  slider publishes `--sidebar-blur` to the document root so the
  same value drives macOS/Windows native vibrancy and Linux CSS
  blur. Light-mode `--background-lightness` dropped 97% → 88% so
  the panel reads as translucent instead of opaque white.
  Faux-glass `@supports` fallback paints noise + radial-light +
  rim-edge for compositors that can't blur.
- **Dedicated AppImage CI job.** New `linux-appimage` job runs in
  parallel with `linux`, builds only the AppImage, uploads as a
  separate artifact. `release` job now needs
  `[linux, linux-appimage, windows]` and includes the AppImage in
  the GitHub Release. Both Linux jobs set `NO_STRIP=1` because the
  Tauri-downloaded `linuxdeploy` AppImage ships an old `strip`
  binary that can't handle the SHT_RELR (0x13) section type
  introduced in glibc 2.36+.

## macOS / Windows

No platform-conditional code outside `tauri.macos.conf.json` (which
restores the macOS-only window options verbatim) and `HARNESS_LABEL`
/ `HARNESS_TITLE`. macOS and Windows builds are unaffected.

## Known gaps (called out for review)

- **Title generation for mcode sessions** uses the literal mcode
  until the user renames the tab. A `mcodeTitle.ts` adapter (~50
  lines) would close this. Same approach as `fxTitle.ts`.
- **One-shot text operations** (commit messages, PR descriptions)
  fall through to `cursor` because `mcode` only ships a live ACP
  path. A `mcodeText.ts` adapter (~80 lines) plus adding
  `"mcode"` to `TEXT_HARNESSES` would close this.
- **Token / cost tracking** doesn't read mcode's usage events yet.
  Needs both an ACP `usage` event handler and UI plumbing.
- **AppImage is 112 MB instead of ~85 MB** because `NO_STRIP=1` is
  on. Proper fix is a tauri-bundler version bump with a newer
  linuxdeploy; deferred.

## Verification

- `npx tsc --noEmit` → **0 errors**
- `cargo check` → **clean**
- `pnpm build:linux:appimage` (earlier in the session) → produced a
  working `MonoCode_0.1.42_amd64.AppImage` (the only path that
  exercises the full Tauri→vite→AppImage chain).

## Checklist

- [x] macOS / Windows code paths not altered
- [x] pnpm install regenerates both pnpm-lock.yaml and
      package-lock.json
- [x] No new top-level dependencies added beyond what was needed
- [x] All new exported functions have docstrings / JSDoc

## Summary by CodeRabbit

- **New Features**
  - Added MiniMax Code support, including model selection, session continuity, prompt handling, and tool activity.
  - Added dedicated Linux AppImage packaging and improved release asset handling.

- **UI Improvements**
  - Updated glass effects across light and dark themes with improved translucency, blur, and fallback rendering.
  - Appearance controls now work in light mode.
  - Refined native window appearance across Windows, macOS, and Linux.

- **Bug Fixes**
  - Prevented an opaque startup frame before glass effects are applied.


hardbeat920:
@sahilcodexx can we split this into multiple prs as it covers different features/fixes. Thank you 🙏

PR #263, opened 2026-09-16
Show OpenCode Go usage in the status-bar footer

D3nnis72:
Shows 5-hour, weekly, and monthly limits with reset countdowns via the official zen/go usage endpoint, reusing UsageProviderChip.

## What changed

When the active session runs on the OpenCode harness, the bottom-left footer now shows an OpenCode Go usage chip — same look as the Codex chip: mini bar with 5h / weekly / monthly percentages plus reset countdowns, and a popover with one progress-bar card per window. No Go subscription → the chip hides itself and the footer looks exactly as before.

## Why

OpenCode Go subscribers had no way to see their plan usage inside the app — the only option was the web console. Claude and Codex already surface this in the footer; OpenCode sessions just showed a plain "opencode" label.

## UI

Before: footer shows a plain `opencode` session label, no usage info.
image

After: footer chip reads `9% 4h 5m · 33% 4d 2h · 28% 24d 20h`; clicking opens the "OpenCode usage" popover with 5-hour / Weekly / Monthly limit cards, each with used %, remaining %, and reset timer. 

image

## Checklist

- [x] I ran `npm run check`

## Summary by CodeRabbit

- **New Features**
  - Added OpenCode Go usage tracking alongside existing providers.
  - OpenCode usage now displays rolling, weekly, and monthly limits when available.
  - Added monthly limit details and reset information to usage views.
  - OpenCode usage refreshes automatically and indicates unavailable accounts appropriately.

- **Bug Fixes**
  - Improved handling of unavailable, unauthorized, malformed, and incomplete usage responses.
  - Improved recognition of OpenCode credentials from supported configuration sources.

- **Tests**
  - Added coverage for OpenCode usage parsing, status handling, and monthly rate-limit data.


hardbeat920:
Thanks for the PR the feature looks great. A few issues should be addressed before merging:

Credential lookup is hardcoded to ~/.local/share/opencode/auth.json, so supported setups using XDG_DATA_HOME, environment variables, or configured API keys will appear disconnected.

Also when Go usage is unavailable, the normal OpenCode session footer disappears and only the refresh button remains.

Usage windows should require an explicit valid status; malformed successful responses should be treated as errors rather than a permanent unavailable state.

Other than that looks good to me

D3nnis72:
Thanks, all three are addressed in 28f34f9:

1. Auth resolution now mirrors OpenCode itself (env blob, then OPENCODE_DATA_DIR, then XDG, then default, plus legacy macOS fallback).
2. Unavailable Go usage falls back to the normal session footer instead of leaving just the refresh button.
3. Windows require explicit ok/rate-limited status. A 200 with no usable windows is a retryable error rather than sticky unavailable.

Checks: 2443 web and 280 Rust tests pass, tsc/fmt/clippy clean.

hardbeat920:
Thanks @D3nnis72, the footer and response-parsing issues look fixed. 

One credential edge case remains: 
configured provider.options.apiKey values still aren’t handled, and a valid OPENCODE_AUTH_CONTENT without opencode-go currently falls back to the disk credentials instead of remaining authoritative like OpenCode does.

D3nnis72:
Alright should be fixed now

PR #289, opened 2026-09-17
fix: sticky file header and full-width diff in inbox PR view

### LINKED ISSUE #265: Diff view (from notifications) isn't full width and file name header doesn't stay sticky while scrolling
### MonoCode version

0.1.49 (also seen on earlier versions)

### Operating system

macOS

### OS version

macOS 26.7 (Tahoe)

### Provider CLI

Multiple / not sure

### What happened

When I open a diff from a notification, the diff page doesn't use the full width of the window — there's wasted space on the sides and the diff is harder to read than it needs to be.

Separately, when scrolling through a long diff, the file name header should stay sticky (pinned to the top) until you've fully scrolled past that file's diff. Right now it scrolls away immediately, so it's easy to lose track of which file you're looking at.

Expected: the diff view uses the full available width, and each file's name header stays pinned while that file's diff is still on screen.

### Steps to reproduce

1. Trigger a notification that includes a diff (e.g. a completed agent turn with file changes).
2. Open the diff from that notification.
3. Notice the diff view doesn't span the full width of the window.
4. Scroll down through a file's diff and notice the file name header doesn't stay pinned/sticky.

MichaelOgunjimi:
Closes #265

The per-file header lost its sticky positioning in the inbox PR diff because the "cards"
layout wrapper set overflow-hidden, which becomes the containing scroll box for position:
sticky and neutralizes it. Move the clipping to a body-only wrapper and give the header its
own rounded corners instead. Also let the Code tab diff use a wider (1600px) column instead
of the 5xl reading width used for the Summary/comments tab.

## What changed

- `UnifiedDiffView.tsx`: header is sticky unconditionally (not just in "stacked" layout);
  moved `overflow-hidden` off the card `<section>` onto a body-only wrapper so it no longer
  breaks the header's sticky positioning.
- `InboxView.tsx`: Code tab diff column now caps at `max-w-[1600px]` instead of `max-w-5xl`.

## Why

Scrolling a long file's diff in the Inbox PR view lost the filename header entirely, and the
diff was squeezed into the same narrow reading-width column used for the PR description.

## UI

**Before** (header scrolls away with content):
image

**After** (header stays pinned while scrolling):
image

## Checklist

- [x] I ran `npm run check`

## Summary by CodeRabbit

- **UI Improvements**
  - Standardized the maximum content width across inbox headers, tabs, and content for a more consistent layout across views.
  - Improved diff view cards with consistent borders, rounded corners, clipped expanded content, and sticky headers in both supported layouts.
  - Applied layout-specific header rounding to create cleaner transitions between card sections and their content.


hardbeat920:
@MichaelOgunjimi i like the sticky header but i am not too sure about the full width view. It looks a bit off to me with the header max-width being diffent... dont you think?

MichaelOgunjimi:
Good catch, and I see exactly why now. It's not the sticky per-file header inside the diff itself — it's one level up. InboxView.tsx has two separate width wrappers:

- The data-inbox-detail-header block (PR title, author, the Summary/Code tab bar) is unconditionally max-w-5xl.
- The tab content body below it is max-w-[1600px] on the Code tab, max-w-5xl on Summary — that's the change this PR made.

So on the Code tab specifically, the title/tab-bar block stays capped at the narrower 5xl while the diff content beneath it widens to 1600px. That's the "header max-width being different" you're seeing — they're just not sharing the same cap.

Fix should be to make the data-inbox-detail-header wrapper follow the same conditional the content wrapper already uses (tab === "code" ? "max-w-[1600px]" : "max-w-5xl"), so the two stay in sync — narrow together on Summary (where 5xl reading-width still makes sense for prose), wide together on Code. I can push that as a follow-up commit on this branch if that sounds right.

hardbeat920:
there are two options we keep the narrower width or we make it full width but i think they should match and if its the full width we should see how that looks on the "Summery" side not just the code changes

MichaelOgunjimi:
I don't think it has to be an either/or between the two globally the fix doesn't force Summary to go wide too. It just makes the header track whichever width is already active for the tab you're on:

- Summary tab: header + content both stay at max-w-5xl, exactly like today. Nothing changes there — PR descriptions and comments keep reading at the same measure they always have.
- Code tab: header + content both widen to max-w-[1600px] together.

So there's nothing to check on the Summary side, since this fix doesn't touch it at all — only the Code tab's header catches up to the width the diff content already got in this PR. Two different, appropriate widths for two different kinds of content (prose vs. code), just consistently applied within each tab instead of the header being stuck at one fixed width regardless of which tab is showing.

Happy to push that as a follow-up commit on this branch if that resolves it.

MichaelOgunjimi:
One correction to my last comment, actually — let's revisit the "keep Summary narrow" part.

Worth being precise about what "full width" means here too: max-w-[1600px] isn't actually 100% fluid width, it's a fixed cap. On a typical 14" laptop screen it happens to fill nearly the whole window (which is why it reads as "full width"), but on a bigger display — a 16" MacBook, an external monitor, anything wider than 1600px — there'd still be visible margin on both sides. It's a wider cap, not true edge-to-edge full width.

Given that, I think it's fine to apply max-w-[1600px] to the Summary tab too, not just Code. Since it's a bounded cap rather than actually stretching prose to fill an ultrawide monitor, it shouldn't read as uncomfortably wide for the PR description/comments either. So: both tabs get max-w-[1600px], header included, everything matching everywhere — simpler than my "match per-tab" suggestion, and it directly answers wanting to see how Summary looks at the wider size.

hardbeat920:
> I don't think it has to be an either/or between the two globally the fix doesn't force Summary to go wide too. It just makes the header track whichever width is already active for the tab you're on:
> 
> * Summary tab: header + content both stay at max-w-5xl, exactly like today. Nothing changes there — PR descriptions and comments keep reading at the same measure they always have.
> * Code tab: header + content both widen to max-w-[1600px] together.
> 
> So there's nothing to check on the Summary side, since this fix doesn't touch it at all — only the Code tab's header catches up to the width the diff content already got in this PR. Two different, appropriate widths for two different kinds of content (prose vs. code), just consistently applied within each tab instead of the header being stuck at one fixed width regardless of which tab is showing.
> 
> Happy to push that as a follow-up commit on this branch if that resolves it.

i am not sure i want to have that layout shift just by changing the tabs it would be awkward

hardbeat920:
> One correction to my last comment, actually — let's revisit the "keep Summary narrow" part.
> 
> Worth being precise about what "full width" means here too: max-w-[1600px] isn't actually 100% fluid width, it's a fixed cap. On a typical 14" laptop screen it happens to fill nearly the whole window (which is why it reads as "full width"), but on a bigger display — a 16" MacBook, an external monitor, anything wider than 1600px — there'd still be visible margin on both sides. It's a wider cap, not true edge-to-edge full width.
> 
> Given that, I think it's fine to apply max-w-[1600px] to the Summary tab too, not just Code. Since it's a bounded cap rather than actually stretching prose to fill an ultrawide monitor, it shouldn't read as uncomfortably wide for the PR description/comments either. So: both tabs get max-w-[1600px], header included, everything matching everywhere — simpler than my "match per-tab" suggestion, and it directly answers wanting to see how Summary looks at the wider size.

can you try and see how that looks

MichaelOgunjimi:
Pushed in [c0d4ee5](https://github.com/hardbeat920/monocode/pull/289/commits/c0d4ee5b2a5c6f695ed3e8bbe4f287d257b9ad6d) @hardbeat920 Update's done 

Header, Summary tab, and Code tab all now share max-w-[1600px] uniformly, no more per-tab or header/content split.

Tried it on an actual ultrawide monitor (3440x1440) to check the exact thing you asked about screenshots below:

1.
image



2.
image



3.
image



4.image

PR #292, opened 2026-09-17
fix: return focus to composer input after a question is answered

### LINKED ISSUE #267: Input/composer field doesn't auto-focus most of the time
### MonoCode version

0.1.49 (also seen on earlier versions)

### Operating system

macOS

### OS version

macOS 26.7 (Tahoe)

### Provider CLI

Multiple / not sure

### What happened

Most of the time, the cursor doesn't automatically land in the message/composer input — for example after switching between sessions/tabs, after the agent finishes a turn, or right after opening/focusing the app. I have to manually click into the input box before I can start typing, every time.

Expected: the composer input should auto-focus in these situations so I can start typing immediately, similar to how most chat/IDE apps behave.

### Steps to reproduce

1. Open MonoCode and switch to a different session/tab.
2. Try typing right away without clicking into the input box first.
3. Notice nothing is typed because the input isn't focused.
4. Also reproducible right after an agent turn finishes, or after switching windows back to MonoCode.

MichaelOgunjimi:
Closes hardbeat920/monocode#267

The issue lists three ways the composer loses focus: after answering a question, after switching windows back to MonoCode, and after an agent turn finishes. All three share the same shape: the composer's DOM-focus effect only re-fires when specific props change, and none of those props changed at the moment focus needed to come back.

## 1. Question answered (`src/chrome/Composer.tsx`)

The focus-restore effect only listed `focused` in its deps:

```tsx
useEffect(() => {
  if (!focused) return;
  if (document.querySelector("[data-model-picker], ...")) return;
  ref.current?.focus();
}, [focused]);
```

`focused` is already `true` before a question appears and never toggles false→true as the question resolves, so the effect never re-ran. Fixed by adding `question` to the deps, so it re-fires when `question` becomes `undefined`.

Also scoped the guard's `document.querySelector` to `ref.current.closest("[data-composer]")` (per CodeRabbit review on an earlier revision of this PR) — inactive inbox `SessionPane`s stay mounted in a hidden container, so a document-wide query could match an unrelated composer's open picker/question-form and wrongly skip focusing the active one.

## 2. Agent turn finishes (`src/chrome/Composer.tsx`)

Same root cause, different trigger: nothing restores focus when the active session's `busy` flag flips from `true` to `false`, in case something else (a clicked Stop/approve button, a click in the transcript) took DOM focus away during the turn without updating the app's `composerFocused` intent. Added `busy` to the same effect's deps so it re-fires on turn completion too.

## 3. Window refocus (`src/App.tsx`, `src/chrome/Composer.tsx`, `src/surfaces/PaneTree.tsx`, `src/surfaces/SessionPane.tsx`)

There's already a Tauri `onFocusChanged` listener, but it only updates the flag used to suppress OS notifications. My first attempt at this fix called `setComposerFocused(true)` there, but CodeRabbit correctly caught that this is a no-op in the common case: `composerFocused` is normally *already* `true` when you alt-tab away, so React bails out of the same-value state update and the Composer's effect never re-runs.

Fixed properly with a monotonic token, mirroring the existing `searchFocusToken`/`focusToken` pattern already used for `SearchView` in this file: a `composerFocusToken` counter in `App.tsx` is bumped alongside `setComposerFocused(true)` on window refocus (only when nothing else already owns focus — `document.activeElement === document.body` and no search/inbox/notes/settings/project-terminal view is open), threaded through `PaneTree` → `SessionPane` → a new `focusToken` prop on `Composer`, and added to the focus-restore effect's deps. A bumped token always looks like a change to the effect, regardless of the previous value.

## 4. Guard against stealing focus from another composer (`src/chrome/Composer.tsx`)

Also caught by CodeRabbit: the picker/question-form guard only checked *this* composer's own subtree. If `document.activeElement` was inside a **different** mounted Composer — e.g. reached by keyboard Tab navigation, which never fires the `onMouseDown`-based `onFocus` that updates which pane is considered "focused" — this composer's effect could still yank focus away from it whenever its own `busy`/`question`/`focusToken` changed. Fixed by comparing `document.activeElement`'s closest `[data-composer]` ancestor against this composer's own and bailing out if they differ.

## 5. Portaled pickers (`src/chrome/Composer.tsx`)

Caught by @hardbeat920 in review: model/access/branch/model-settings/file pickers render through `Popover.tsx`'s `createPortal(..., document.body)`, so they never live under a composer's own `[data-composer]` subtree. The scoping from fix #1 made the guard blind to those, so `busy`/`question`/`focusToken` changes could steal focus back to the textarea while one was open. Split the check: inline controls (skill picker, session folder picker, mention picker, composer-plus, question form) stay scoped to the composer's subtree; portaled pickers are checked document-wide since they have no subtree to scope to.

## Testing

- `npm run check` (vitest, tsc, cargo fmt/clippy/test) all pass.
- `src/chrome/Composer.test.ts` covers all five fixes by mounting the real `Composer` and asserting `document.activeElement`: question-answered, turn-finished, a token bump while `focused` stays `true` throughout, a control inside a different composer keeping its focus, and a portaled picker keeping its focus. Verified each new test fails without its corresponding fix.


hardbeat920:
thank you @MichaelOgunjimi overall looks great just one small thing

This scoped query misses portaled pickers rendered under document.body. When busy changes, focus can be stolen from an open picker and returned to the textarea. 

Could we guard portaled surfaces separately and add a regression test?

MichaelOgunjimi:
Good catch — the `[data-composer]`-scoped query missed model/access/branch/model-settings/file pickers, since `Popover.tsx` renders those through `createPortal(..., document.body)`, outside the composer's own subtree.

Fixed in 1b1b0ac: inline controls (skill picker, session folder picker, mention picker, composer-plus, question form) stay scoped to the composer's subtree; the portaled pickers are now checked document-wide separately, since they have no subtree to scope to. Added a regression test in `Composer.test.ts` that simulates a portaled picker's content and verified it fails without the fix.

hardbeat920:
@MichaelOgunjimi thank you merged :)

Issue #308, opened 2026-09-17
A live turn stops reporting progress while the agent thinks, once its work group is closed


yankawai:
### MonoCode version

0.1.50

### Operating system

macOS

### OS version

macOS 26.5.2

### Provider CLI

Claude Code

### What happened

While a turn is running, the only full-size text in it is the last assistant prose. Claude Code narrates in reasoning: one short preamble at the start of the turn, then `thinking` between tool calls. Those thoughts render as one-line steps inside the activity group, so once that group is closed the turn stops reporting anything. It closes two ways: I collapse it myself to keep a long transcript readable, and it also folds once the agent has started answering.

On a 30 minute turn here (Fable 5.1, 93 tool calls) the only line on screen was the sentence the model wrote in its first second. The same session's stream carried 26 `thinking` blocks and 1 `text` block, so there was plenty to show — all of it behind the closed group.

Codex sessions do not read this way: that CLI narrates in `agent_message` prose, which lands outside the group as full-size text and keeps arriving through the turn.

Expected: a closed live group still says what the agent is on now, the way the thinking row inside it already does — "the line that keeps a long think from reading as a stall".

### Steps to reproduce

1. Start a Claude Code session on a task that runs tools for several minutes.
2. When the first activity group appears, click its header to close it.
3. Watch the turn: the header keeps counting calls, and nothing else changes until the agent writes prose again — with Claude Code that can be the end of the turn.

yankawai:
Fix in #309: the closed live group keeps the newest thought on one line under its header.

yankawai:
Understood on #309 — showing reasoning under a header the reader collapsed was the wrong shape for it. I went back through the code and the history before trying again, and the picture is narrower and worse than what I wrote when I opened this: a live turn driven by a reasoning-narrating harness has no progress line at all, and it used to have one.

**MonoCode had this line until 0.1.18.** The live turn rendered a ticker — `src/hooks/useActivityTicker.ts` plus `ActivityTicker` in `AgentTranscript.tsx` — and it held the newest step whatever it happened to be:

```ts
function ActivityRow({ block, ... }) {
  if (isThinkingBlock(block)) return <ActivityThinkingRow ... />;  // a thought could hold the line
  if (isProseBlock(block))    return <ActivityNoteRow ... />;
  return <ActivityToolRow ... />;
}
```

Three commits took it away, none of them intending to:

| commit | release | what it did |
|---|---|---|
[...]

PR #309, opened 2026-09-17
Keep the newest thought on a closed live group

yankawai:
## What changed

A live activity group that is closed keeps the agent's newest thought on one line under its header. `latestThinkingStep` in `transcriptActivity.ts` picks it out; `ActivityPhaseGroup` renders it with the same `ActivityThinkingRow` an open group uses. An open group, a settled turn, and a phase with no thoughts are untouched.

## Why

Closing a group folds its steps away, and with them the thinking row that is there to keep a long think from reading as a stall. Harnesses that narrate in reasoning rather than prose hit that hardest: Claude Code writes one preamble and then thinks between calls, so a 30 minute turn here showed nothing but the sentence written in its first second. Details in #308.

## UI

Closed live group, before:

```
🔍 Explored the project · Running 93 commands · Edited 3 files
```

after:

```
🔍 Explored the project · Running 93 commands · Edited 3 files
   Pulling the frames out of the recording.
```

The second line is the existing dim thinking row, truncated to one line, and it follows the agent as it moves on.

## Checklist

- [x] I ran `npm run check`

`npm run check:web`: `tsc --noEmit` is clean and the transcript suites pass, including three new cases for this row. A set of localStorage-dependent tests fails identically on an unmodified `main` on this machine (Node 26 wants `--localstorage-file`), and `check:rust` stops on `clippy::nonminimal_bool` at `src-tauri/src/control.rs:290` under clippy 1.96 — also on unmodified `main`. This PR touches no Rust.

## Summary by CodeRabbit

* **New Features**
  * Ongoing reasoning now remains visible beneath a collapsed activity group.
  * The displayed reasoning updates as new thoughts are generated and clears when the turn is complete.

* **Tests**
  * Added coverage for showing the latest reasoning step, skipping empty entries, and handling turns without reasoning.


hardbeat920:
Thanks for working on this @yankawai I don’t think we want to show reasoning after the activity group has been collapsed, so I’m going to close this for now. 

We can revisit the underlying progress-feedback issue with a different approach in the future if you dont mind.

Thank you again, i really appreciate your work 🙏

yankawai:
Took the different approach you suggested: #318. It leaves a collapsed group alone — the header of a live, open group follows the newest thought, and the count comes back the moment the group is closed or the turn settles. What led me there is in #308: the live line existed until 0.1.18 and was lost to the group redesign for harnesses that narrate in reasoning.

PR #254, opened 2026-09-16
Add Kimi Code and Antigravity ACP harnesses

elijah7x:
## Summary
- Two new live agent harnesses, `kimi` and `antigravity`, joining the ACP family alongside cursor/grok/fx. Spawn targets: `kimi acp` and `agy_acp_server.par`; both verified end-to-end (`initialize` → `session/new` → `session/prompt` → `end_turn`) against the real servers.
- Real permission flow rather than fx's blanket auto-approve: `session/request_permission` surfaces in the approval UI and resolves by advertised option kind; cancel releases pending approvals with the ACP `cancelled` outcome. Mode downgrades fail closed so a rejected `session/set_mode` cannot leave a looser mode active.
- Rust resolvers prefer `~/.kimi-code/bin/kimi` (guarded against the legacy Python `kimi-cli`) and the `agy_acp_server.par` wrapper that sets the server's resource directory; both binaries join orphan-process recognition.
- Model catalogs probe ACP `initialize` + `session/new` `configOptions` for live values (incl. Antigravity's top-level model list), with seed models covering the offline case.
- Official brand icons extracted from the installed apps; skills discovery gains `.kimi-code/skills` (project + `KIMI_CODE_HOME` user root) and `~/.gemini/antigravity/skills`.

#### Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run` — 224 files / 2438 tests (incl. new offline ACP transport suite: spawn args, approvals, plan-mode denial, cancel, resume, catalog probe, auth errors)
- [x] `cargo fmt --check`, `cargo clippy -D warnings`, `cargo test` — 278 tests
- [x] Live ACP regression: real `session/prompt` round-trip on both `kimi acp` 0.43.1 and `agy_acp_server.par`
- [ ] Manual smoke in the desktop app: pick each provider in the model picker, send a turn, exercise approve/deny and cancel

Generated with [Devin](https://devin.ai)

## Summary by CodeRabbit

- **New Features**
  - Added Kimi Code and Antigravity as supported AI harnesses.
  - Added model selection, live availability detection, model catalog discovery, and provider-specific icons.
  - Added session controls including resume, cancellation, approvals, and session cleanup.
  - Added support for prompts with attachments, planning modes, runtime settings, and tool/event updates.
  - Added discovery of Kimi Code and Antigravity skills from project and user locations.
- **Tests**
  - Expanded coverage for provider detection, sessions, permissions, model discovery, and skill loading.

PR #314, opened 2026-09-18
Add Antigravity ACP provider

elijah7x:
Split out from #254 as requested — this PR is Antigravity only; Kimi Code comes separately.

Antigravity speaks Agent Client Protocol over stdio via the `agy_acp_server.par` wrapper, joining the existing ACP provider family. It surfaces real permission requests to the approval UI, matched by advertised option kind; mode downgrades fail closed so a rejected `session/set_mode` cannot leave yolo active.

- Rust resolver prefers `~/.local/bin/agy_acp_server.par` so the server's resource directory is set correctly.
- Catalog probes `initialize` + `session/new` `configOptions` so the model picker reflects live provider values; seed models cover offline.
- Skills discovery gains `~/.gemini/antigravity/skills`.

## Transport and lifecycle hardening

Testing against the real server exposed a family of lifecycle defects that could leave a turn stuck "thinking" forever — e.g. when `session/prompt` never resolves because a server-side post-turn hook wedges, or when a cancelled turn's transport gets reused. This PR hardens the shared ACP path (`jsonRpc.ts`, `child.ts`, `harness.rs`) along with the provider:

- Generation-scoped child ids (`thread#N`) isolate output from retired processes so late stdout/exit can never reach a replacement client.
- `sessionEpoch`/`cancelEpoch` invalidate queued lifecycle work on cancel/stop/forget; startup re-checks retirement after binary resolution and before publishing, so a forgotten session can never resurrect a process.
- Turns are serialized per thread; `turnActive` spans model/config/mode selection through the prompt, and a mid-flight cancel marks the transport stale so the next send recycles and resumes via `session/resume`.
- JSON-RPC writes race a 15s deadline; cancellation settles local pending requests before notifying the provider; permission-response write failures retire the generation instead of leaving the prompt parked.
- A silence watchdog surfaces a status hint on long quiet turns while the 30-minute prompt timeout recycles the generation.
- `harness_write` runs on the blocking thread pool via `spawn_blocking` so a child that stops draining stdin cannot starve the async workers dispatching `harness_kill`.

#### Test plan

- [x] Mocked lifecycle suite (`antigravityLive.test.ts`): 36 cases covering queued-send suppression, forget resurrection, config-phase cancel, blocked writes, approval write failures, resume/load/new ladder, stopReason handling
- [x] `jsonRpc.test.ts`: write deadlines, pending-request cleanup, no unhandled rejections
- [x] `child.test.ts`: generation-keyed routing through the real Tauri event mock
- [x] Real-binary suite (`antigravityReal.test.ts`, `AGY_REAL=1`): end-to-end turn + transport reuse, mid-prompt cancel + recovery, process exit + resume — 3/3 pass
- [x] Soak suite (`antigravitySoak.test.ts`, `AGY_SOAK=1`): 10 scenarios against the real server — 6-thread fleets, rapid queued sends, 5 mid-permission cancel cycles, stop/forget cycles, mid-turn SIGKILLs, cwd drift, queued-send suppression, concurrent mixed abuse, long-task interrupt+continue — 10/10 pass
- [x] `vitest` full suite, `tsc`, `cargo test`, `cargo fmt --check`, `vite build` all clean

## Summary by CodeRabbit

- **New Features**
  - Added Antigravity as a supported AI harness with model discovery, session management, prompt execution, cancellation, approvals, and recovery.
  - Added Antigravity to the model picker with provider branding and a default model.
  - Added discovery of user skills from Antigravity’s skills directory.

- **Bug Fixes**
  - Improved recovery from stalled process communication and blocked writes.
  - Ensured cancelled or replaced sessions do not receive late process events.

- **Tests**
  - Added comprehensive coverage for Antigravity sessions, permissions, model discovery, cancellation, recovery, and concurrent usage.


hardbeat920:
Nice work overall @elijah7x 

Before merging, could we address two gaps, advertise ACP session.configOptions before relying on session/set_config_option, and ensure Stop cancels a generation still inside initialize/session setup. Currently setup cancellation can block the next send for up to 45 seconds. Also, please replace the hardcoded /Users/elijah/... test path with an AGY_BIN override plus home-directory fallback.

Thank you again 🙏

elijah7x:
Thanks for the review! All three points are addressed in the latest push:

1. **configOptions gating** — `setConfigOption` now only issues `session/set_config_option` when the session actually advertised that option id via `configOptions` (from `session/new`/`load`/`resume`, kept fresh by `config_option_update` notifications). An unadvertised id is skipped rather than sent. Model settings were already gated through `resolveSettingConfigId`; the model selector now follows the same rule. New tests cover a session advertising zero options and one advertising a subset.

2. **Stop/cancel during initialize/session setup** — startups that have spawned but not yet published their `Live` now register in a `pendingSetupByThread` map. `cancel`/`stop`/`forget` reject the in-flight wire calls (`initialize`, `session/new`, `session/resume`, `session/load`) and kill that generation's child immediately, so the next send no longer waits out the 45s session timeout. A `session.ended` from a generation retired by user intent is suppressed. Covered by a new test that parks `session/new` mid-flight and cancels/stops.

3. **Binary path override** — both `antigravityReal.test.ts` and `antigravitySoak.test.ts` now resolve `BIN` from `AGY_BIN` with a `~/.local/bin/agy_acp_server.par` fallback via `os.homedir()`, and the `../fs` mock returns `homedir()`.

Also folded in the review-bot findings while here: `hasUsageFields` now recognizes every key `usageFromUpdate` parses (snake_case tokens, cache and context-window fields), and `resolve_antigravity` returns `None` on Windows since the `.par` wrapper is POSIX-only.

The same fixes were applied symmetrically to the Kimi PR (#315) where the code paths are shared (`KIMI_BIN` override there).

hardbeat920:
Thanks @elijah7x, the latest changes look good overall. Two small things still stood out:

1. We still don’t advertise session.configOptions, so a compliant ACP server may omit them and the selected model would be silently ignored.
2. If cancellation happens during session/resume, it can still fall through to session/load and potentially wait for the 45-second timeout.

Could we check the retired state before each resume/load/new fallback, or close the pending ACP client when setup is cancelled?

elijah7x:
Thanks — both points addressed in the latest push (agy `893b650`, kimi `a38cb01`).

1. **Advertising `session.configOptions`**: `clientCapabilities.session.configOptions = { boolean: {} }` is now sent in `initialize` — both in the live adapter and the catalog-probe connection, since the probe also relies on `configOptions` to enumerate models. Per the schema, supplying `{ boolean: {} }` also lets agents emit `type: "boolean"` options, so `set_config_option` now encodes them properly: `{ type: "boolean", value: <bool> }` for boolean-typed options, and the plain `value` string (value-id variant) otherwise. `readConfigOptions` preserves each option's `type` from `session/new`/`load`/`resume` and `config_option_update` notifications.

2. **Cancel during `session/resume`**: did both of your suggestions — the pending-setup abort now `close()`es the ACP client (in-flight requests reject and any later request fails immediately), and `retired()` checks guard `session/load` and `session/new`, so a cancelled setup can never continue down the fallback ladder into another 45s wait. A cancelled setup also can't publish `liveByThread`/`resumeByThread` or emit session events.

New regression coverage on both providers: cancel landing mid-`session/resume` issues no `session/load` or `session/new`; `initialize` advertises `session.configOptions`; boolean options go out typed; a cancel reentered from the `session.started` listener can't orphan the just-published transport.

Checks: focused lifecycle suites 39/39 each, `tsc --noEmit` clean, real-binary smoke 3/3 each (turn e2e + transport reuse, mid-prompt cancel + recovery, provider kill + respawn/resume), full suite green on CI.

hardbeat920:
@elijah7x thanks for this, people were waiting for it 🙏

PR #324, opened 2026-09-19
Add `amend last commit` to the Changes panel

kartava:
## What changed

The Changes panel can now amend the last commit. Open the commit menu and pick **Amend Last Commit**. The message box fills with the current HEAD message, and you can reword even with nothing staged.

If the commit is already pushed, MonoCode warns you first, since it has no force push.

## Why

Fixing a typo or adding a forgotten file to the last commit meant leaving MonoCode for the terminal. This keeps that small, common task in the app.

## UI

New item at the bottom of the commit options menu.
image

## Checklist

- [x] I ran `npm run check`

## Summary by CodeRabbit

- **New Features**
  - Added support for amending the latest Git commit, including editing its message.
  - Added retrieval of the current commit’s full message, including subject and body.
  - Amended commits can be created without staged files, and the interface resets after committing.
  - Added accurate detection of whether the current commit has been pushed, including branches without configured upstreams.

- **Bug Fixes**
  - Amend mode now resets when the selected branch or commit changes.


hardbeat920:
Looks good overall @kartava 

One concern, if amend mode is enabled and the user switches branches, the old amend message remains and could amend the wrong commit. 

Could we reset amend mode and the message whenever the branch or HEAD changes?

kartava:
@hardbeat920 Done.

hardbeat920:
Sorry, I missed one detail in my earlier review. The HEAD-change handling looks good, but two branches can point to the same HEAD. In that case, switching branches would keep amend mode active. Could we also reset it when the branch name changes?

kartava:
@hardbeat920 Done.

PR #325, opened 2026-09-19
Fix Add to chat when no session tab is available

### LINKED ISSUE #311: Add to chat doesn't work when no session tab is available
### MonoCode version

0.1.50

### Operating system

macOS

### OS version

26.6.2

### Provider CLI

Claude Code

### What happened

When clicking add to chat on diff view, it does nothing when no session tab is opened

Also the active tab is not draggable, only non-active tabs are draggable

### Steps to reproduce

1. Close session tab
2. Click add to chat on a source file

stillnovice:
Fixes #311.

## Changes

**src/App.tsx — `openSessionForAddToChat`**: when the workspace has zero tabs, the handler silently returned and the request was dropped. Build a fallback tab + session instead: seed via `newDefaultSession` (harness/model/modelSettings from `sessionsRef.current[0]`, mirroring the `onCloseAllTabs` pattern), wrap in `newTab`, continue the existing split-pane flow. All other early-returns unchanged.

**src/surfaces/SessionPane.tsx — seed precedence**: `initialDraft` chain is now `draftRef.current ?? (inbox/note/handoff card ? undefined : composerSeed ?? restoredDraft)`. A persisted draft for a reused session id can no longer clobber the one-shot seed. Card exclusion and `draftRef`-wins-for-live-typing preserved.

## Tests

- `workspaceTabGroups.test.ts`: new case — tab whose layout holds only unmapped leaf ids still splits correctly.
- `SessionPane.test.ts` (new): seed beats a persisted draft (`composer_draft_get` resolving text); no-seed falls back to persisted draft.

## Verification

- `npx vitest run` → 244 files / 2628 tests pass (pre-rebase; rebased onto latest main)
- `npx tsc --noEmit` → clean (re-run post-rebase)
- cargo untouched

## Summary by CodeRabbit

- **Bug Fixes**
  - Adding a chat now works even when all workspace tabs are closed by automatically creating a session and workspace tab.
  - New chats preserve the selected harness, model, settings, runtime mode, and working directory.
  - Newly added chats are focused correctly and can open beside existing workspace content without being blocked.
  - Diff focus is cleared when a new chat is added.
- **Tests**
  - Added regression coverage for empty-workspace chat creation and workspace-pane behavior.


stillnovice:
Rebased onto main (c4acf48). Conflict resolution dropped the 6 stacked `composer draft` persistence commits — main shipped a different draft design (explicit draft messages in `is_draft` column, bf991f2), which also makes the seed-vs-draft precedence fix obsolete (no auto-persisted draft can clobber a seed). PR now contains only the zero-tab fix for `openSessionForAddToChat` + the unmapped-leaves test. Full suite: 2679 tests pass, tsc clean.

hardbeat920:
Thanks for the fix @stillNovice  I noticed two issues with the zero-tab path:

1. It creates two chat sessions. The first is an empty fallback session, while the quoted text is added to a second session. This also means the preserved harness and model settings belong to the empty session. Could we create one session with the composer seed and use it directly for the new tab?
2. When there is no active tab, sessionDefaults comes from sessions[0]. That session may belong to another project, so the new chat could use the wrong working directory. I think the fallback should prefer projectCwdRef.current.

The new test only covers openAddToChatSessionPane, which was not changed by this PR. It would be helpful to add a test for the actual tabs = [] flow and confirm that one correctly configured session is created.

stillnovice:
Addressed review points @hardbeat920:

**1. One session, not two** — the zero-tab path now builds exactly one session carrying the composer seed and uses it directly as the fallback tab's pane, so its harness/model/settings are the ones the new chat actually gets. Extracted the flow into a pure helper `addChatToEmptyWorkspace` (src/lib/addChatToEmptyWorkspace.ts) to make it testable.

**2. cwd fallback** — the fallback session's cwd is always `projectCwdRef.current`; no session-derived cwd is consulted, so a donor from another project can't leak its working directory. Note: also switched the settings donor to the *last* known session (sessions[0] could otherwise donate settings cross-project once its cwd no longer wins). Happy to revert to [0] if you'd rather.

**3. Tests** — new `addChatToEmptyWorkspace.test.ts` covers the actual `tabs = []` flow: exactly one session + one tab, tab pane = session id, composer seed present, donor harness/model/runtimeMode preserved, project cwd beats donor cwd, last-session donor preference, Claude defaults with empty workspace, whitespace-only text rejected.

`npx vitest run` → 245 files / 2686 tests pass; `npx tsc --noEmit` clean.

hardbeat920:
Thanks for the updates, @stillnovice. The cwd handling looks fixed, but I found one remaining issue.

The fallback helper already creates a tab containing the new session. `App` then passes that tab and the same session ID to `openAddToChatSessionPane`, which splits the session beside itself. This creates two panes with the same session ID. Could we use the helper’s tab directly when `createdSession` exists, and only call `openAddToChatSessionPane` for existing file-only tabs?

It would also be good to test the final state after the `App` integration, since the current helper tests stop before the duplicate split happens.

Thank you again for your work

stillnovice:
Addressed in 36043fe:

**Duplicate split fixed** — when the zero-tab fallback creates the session, its tab is used directly as the final state. `openAddToChatSessionPane` is now only reached for existing file-only tabs, so the new session can never be split beside itself.

**Integration coverage** — went one step further and extracted the entire handler into a pure function, `applyAddToChatRequest` (src/features/sessions/model/addChatToWorkspace.ts). `App` now just applies the returned `{ sessions, tabs, activeTabId }` — zero logic left in the effect. Tests cover the final integrated state for both paths:

- zero-tab: exactly one session + one tab, tab layout = `[sessionId]` (no duplicate panes), composer seed present, donor harness/model/runtime preserved, project cwd beats donor cwd, last-session donor, Claude defaults on empty workspace
- file-only tab: new session split beside the file pane, focused, no bail
- mounted-session tab: bails (normal path owns it)

`npx vitest run` → 257 files / 3002 tests pass; `npx tsc --noEmit` clean.

hardbeat920:
@stillnovice thank you its ready to merge

PR #336, opened 2026-09-21
fix: keep composer draft when a session's pane closes and reopens

### LINKED ISSUE #270: [Bug]: Composer drafts aren't persisted — silently lost on reload or session close
### Problem

MonoCode's chat composer holds what you type as a plain in-memory React ref (`draftRef` in `src/surfaces/SessionPane.tsx`), not anything persisted. There's no disk write, no Tauri `invoke`, nothing. There's a `composerSeed` field declared on the `Session` type (`src/lib/session.ts`) that looks like it's meant to carry a draft forward, but nothing in the codebase ever assigns it — it's dead code, always `undefined`.

Practical effect: start typing a message, then reload the app (including via the new Cmd+Shift+R / Command Palette reload from #269), close the session pane, or just restart MonoCode, and the draft is gone with zero warning. It isn't hidden somewhere hard to find — it was never written down.

Worth noting: the reload confirmation added for #269/#296 (`confirmReload` in `src/lib/appLifecycle.ts`) only checks `dirtyFiles` (unsaved edits to open file-editor tabs), so it doesn't cover this at all — reloading with a half-typed message in the composer gives no warning either.

### Proposed solution

Actually persist the composer draft per-session — e.g. debounce-write it to disk (or otherwise durable storage) keyed by session id, and load it back into `initialDraft` when the session pane mounts. Once that exists, it would also be reasonable to extend the reload confirmation to cover an in-progress draft, not just dirty files.

### Alternatives considered

None that solve the actual loss — you just have to remember to send or manually copy out anything you don't want to lose before reloading or closing a session.

### Additional context

Originally filed this as a discoverability issue ("can't find where autosaved drafts live"), assuming MonoCode was saving them somewhere I couldn't locate. After digging into the code, that assumption was wrong — there's no persistence path at all, so this is a data-loss bug rather than a missing "reveal in Finder" button.

bluzername:
Fixes #270.

## What was happening

Composer text was only kept in `draftRef`, a plain React ref local to one
`SessionPane` instance (`src/features/sessions/ui/SessionPane.tsx`). That
ref resets to `undefined` every time `SessionPane` unmounts. Closing a
session's pane (or moving it to another split) unmounts it, so whatever you
had typed and not sent was gone, with no warning, the next time you opened
that session.

`composerSeed` on `Session` looked like it was meant to carry the draft
forward, but nothing ever wrote to it from typing - it stayed `undefined`
always, matching what #270 already found.

## What this changes

A small module-level cache, `src/features/sessions/model/draftCache.ts`,
keyed by session id. It lives outside the React tree, so it survives a
`SessionPane` instance being unmounted and remounted, for as long as the
app process keeps running. `SessionPane` now seeds `draftRef` from it on
mount and writes to it in the same place it already updates `draftRef`
(the `onDraftChange` handler passed to `Composer`).

## What this does not fix

It does not make a draft survive a full app restart or the Cmd+Shift+R
reload mentioned in the issue. That needs the draft written into the
session's own persisted record on disk (`session_upsert` on the Rust
side), which is a bigger, separate change across both the TS and Rust
layers - happy to look at that as a follow-up if it's wanted, but wanted
to keep this PR to one focused fix as CONTRIBUTING.md asks.

It does fix the more everyday case: type a message, close the tab (or
switch it into another split, or just click away in a flow that unmounts
the pane) and reopen the session - the draft is there again.

## Testing

`src/features/sessions/model/draftCache.test.ts` - RED before this change
(the module did not exist, so the import failed), GREEN after. Also ran
the full `npx vitest run` (3008 tests, all pass) and `npx tsc --noEmit`
(clean) on the branch.

## Summary by CodeRabbit

* **New Features**
  * Composer drafts are now preserved when switching between sessions during an app session.
  * Clearing the composer removes the saved draft for that session.
  * Drafts remain isolated between sessions.

* **Bug Fixes**
  * Restores each session’s previously entered composer text when revisiting it.

* **Tests**
  * Added coverage for draft persistence, isolation, clearing, and empty drafts.


hardbeat920:
Thanks @bluzername! 

This looks good for preserving drafts across pane close/reopen. Since reload and restart persistence is clearly documented as follow-up work, no blocker from me. 

One small request, change Fixes #270 to Refs #270 so the broader persistence issue stays open.

bluzername:
Thanks @hardbeat920. I change it to Refs #270 so that issue stay open for the reload/restart case.

hardbeat920:
Thanks @bluzername, appreciate the update! Looks good to me.

PR #343, opened 2026-09-21
feat(harness): add OpenCrabs provider (ACP over stdio)

moneyacademyKE:
# OpenCrabs harness (ACP over stdio)

## What

Adds **OpenCrabs** as a harness provider (crab icon in the picker), driven through MonoCode's existing ACP client machinery.

OpenCrabs ships a native ACP server mode (`opencrabs acp` over stdio) — merged upstream as [adolfousier/opencrabs#1540](https://github.com/adolfousier/opencrabs/pull/1540), **released in v0.5.2**. This adapter is live code against a released binary: with the binary installed, the availability probe surfaces the provider automatically; without it, the provider stays hidden and nothing else changes.

## Shape

Ported to the current provider layout — 9 modules under `src/integrations/harness/providers/opencrabs/` (~1,570 LOC + colocated tests), following the fx/Cursor template:

| Piece | What |
|---|---|
| `opencrabs.ts` + `opencrabsProtocol.ts` | Spawns `opencrabs acp`, speaks ACP over stdio via the shared `AcpClient`; full lifecycle: `session/new` / `session/load` (cross-process resume), streamed prompts, tool-call rendering, plan/usage updates, cancel, interactive approvals |
| `opencrabsCommands.ts` | ACP-native command discovery: the server pushes `available_commands_update` at session creation (built-ins, skills, user commands) → the command picker |
| `opencrabsText.ts` / `opencrabsGit.ts` | Session titles + commit/PR/branch text generation via `opencrabs run --quiet --format json` |
| `opencrabsPrompt.ts` | Image/audio/file attachments → `resource_link` (pasted blobs persisted through the shared attachment command) |
| `src-tauri` `harness_resolve_opencrabs` | Binary resolution: `~/.opencrabs/bin/opencrabs` first, then PATH — local builds override the installed release |

Plus: model catalog overlay (provider/model pairs advertised at session start, switchable mid-session), native `session/set_mode` (plan mode enforced server-side), `session/compact`, and `configChanged` emission on model switches.

Design rule throughout: anything the server doesn't answer fails **loud** with a named error — a session never hangs on a silent "Working…".

## Verification

- `tsc --noEmit` clean
- vitest: **13/13** adapter tests (prompt-block mapping, one-shot text runner)
- Rust side: `cargo check` clean
- Rebased on current `main` (v0.1.53) with zero conflicts
- Live wire smoke of the exact ACP contract this adapter speaks: handshake → session resume across processes → `-32601` on unknown methods → streamed turn with tool calls ending `stopReason: "end_turn"`; GUI smoke-tested end-to-end against a live binary

## On the provider pause

CONTRIBUTING asks to hold new harness PRs while the current providers converge on shared patterns. I'm sending now because the server side shipped and this is live against a release rather than a promise — but happy to hold, rebase onto whatever consolidated pattern you land on, or close on request. The branch tracks `main` and stays current either way.

## Summary by CodeRabbit

* **New Features**
  * Added OpenCrabs as a supported harness.
  * Added OpenCrabs availability detection and installation guidance.
  * Added session management, streaming turns, approvals, cancellation, context compaction, and native slash commands.
  * Added support for text and file attachments, including images and audio.
  * Added OpenCrabs-powered session titles, commit messages, pull request content, and branch names.
  * Added OpenCrabs model selection and provider branding.
* **Tests**
  * Added coverage for attachment handling and text generation.


moneyacademyKE:
Both findings fixed in 27e1e63:

**Timeout path (major):** confirmed the mechanism — `killChild` calls `unwatchChild`, so the exit watcher is gone before the Tauri exit event arrives and `await exitPromise` never resolves, wedging the serialized turns queue. Fix follows the suggested shape: `notifyExit()` before the kill (resolves the promise), the timeout kill is awaited in `finally` instead of issuing a second kill, and a timed-out run now rejects with an explicit error before summary parsing. Regression test covers all three properties: rejection, exactly one kill, queue usable on the next call.

**Install strings:** both now point at `adolfousier/opencrabs`.

Verified: `tsc --noEmit` clean, adapter suite 14/14.

moneyacademyKE:
Round-3 audit follow-up: five interaction fixes for how the adapter operates against released opencrabs binaries (a8c0f8e).

1. **Steer was a no-op on every released binary.** The adapter sent only the ext-prefixed `_session/steer` notify, but v0.5.2 registers only the plain `session/steer` — and JSON-RPC drops unknown notifications silently, so mid-turn messages vanished without an error. Now sends the plain name (the parity server accepts both spellings).

2. **Approval dialogs outlived the server's permission timeout.** The server abandons an unanswered `session/request_permission` after 300s and proceeds as denied; the MonoCode dialog stayed open forever, and answering it later sent a response nobody waited for. Dialogs now expire at 290s — just under the server deadline — resolving deny so UI and server agree. Settle-once guarantees no double resolution.

3. **A failed compact wedged the session.** `compactOpenCrabsContext` propagated failures without recycling the transport, unlike prompt turns — one timed-out compact left every later turn on a dead process. It now recycles exactly like a failed prompt turn (regression-tested: failed compact → child killed → next compact respawns).

4. **Failed model switches were silent.** A failed `session/set_model` left the picker showing the new model while the turn ran on the old one. It now emits a `session.error` (the same non-fatal channel fx uses) naming the failed pick and the fallback.

5. **Stale availability hint.** "May not be released yet" — ACP shipped in v0.5.2. The help text now states the real minimum and upgrade path.

Verification: tsc clean, 17/17 provider tests (8 prompt, 2 new approval-expiry, 1 new compact-recycle, 6 text). The doubled-answer fix from the same audit is server-side — on the opencrabs parity branch.

moneyacademyKE:
Follow-up fixes from behavioral audit rounds 3 and 4 (`ef7a486`, `1edb829`) — all verified: tsc clean, 20/20 adapter tests.

**Resume window honesty** (`ef7a486`): two bugs in the session/load fallback path.
- A failed resume silently fell back to a fresh session — transcript looked continuous, but the agent had amnesia for everything above the fold. Now emits an `interjection` at the boundary naming the failure reason, so context loss is visible.
- The `available_commands_update` push lands inside the transcript-replay mute window and before the live session exists — both gates dropped it, so every restarted session lost its autocomplete catalog. The push is now parsed and cached before the mute gates: muting keeps replay out of the transcript, not control-plane data out of the cache.

**Context meter was dead for every session** (`1edb829`): the server emits the ACP-spec usage shape `{used, size}`; the parser read only `window`/`contextWindow`/`context_window` — `size` matched nothing, `contextRatio()` returned null, and the meter rendered nothing, always. `size` is now a window source, pinned by tests against the server's actual wire shape.

Also verified clean this round: title/git generation error paths (loud failures with fallbacks), attachment persistence (pid-nanos collision-proof paths, size cap), multi-session concurrency (process per thread), and the steer race window (messages are delayed to the next turn, never lost — enhancement candidate: auto-flush when idle).

moneyacademyKE:
Both findings fixed in `042652f`, verified (tsc clean, 23/23 adapter tests):

**Interjection timing:** confirmed — the notice fired inside the `session/load` catch, before the fallback `session/new` ran, so a failed fallback would leave the UI claiming a session that never existed. The failure reason is now stored and the notice emitted only after a valid session id is confirmed.

**usage.size validation:** confirmed — `numberField` accepted negative/fractional values, so protocol garbage could replace the meter window. `acpSizeField` now accepts only non-negative integers; the other window field fallbacks are unchanged. Three new tests pin accept/reject behavior.

moneyacademyKE:
**fix: omit --model for the Default catalog placeholder** (8da3d7c)

Wire-proven during end-to-end dogfooding: every child spawned with `--model default` because the static catalog's Default entry leaked its placeholder id into the spawn args — it only worked because the server falls back to its configured model on an unknown id. The flag is now omitted for the placeholder (case-insensitive), so the server default is used deliberately; real pairs still pass through whole (`set_model` is pair-aware).

Verified in the same dogfood session: post-fix, the session footer reports the server's real default model instead of the placeholder. Two new tests pin both behaviors; provider suite 25/25, tsc clean.

hardbeat920:
@moneyacademyKE as mentioned in your previous pr i am trying to be cautious adding too many harnesses. I will keep this as draft until we re-evaluate. Thank you 🙏

